PR #8, opened 2019-06-13
Create minified JS and CSS at 'build' time

jamescrowley:
We can't easily generate these on the fly in Google AppEngine environment, so generate them at build time instead.

We have no hashes or cache busting in place which isn't great, but can still take the manual approach that was used before. 

Ultimately should be switching to webpack etc but avoiding adopting too much tooling when we don't know the future lifetime of this code base yet, and safest just to re-use the existing code that was in place.


jamescrowley:
(Don't merge yet, it looks like the bundling has broken somewhere given minified JS size differences)

PR #18, opened 2019-06-20
Consolidate error reporting so Sentry can report these

jamescrowley:
Consolidate error reporting so Sentry can report these, and we show a error landing page even for code failures early in the loading of the library stack

Lots of warnings are currently being supressed but can't easily resolve them,
so decided to preserve that behaviour.

Notes:

- now forcing the whole app to run through a single entry point again
to simplify where we hook up error handling. Hence including the htaccess necessary
for this.
- if you're running from docker you'll need to do `docker-compose build` to rebuild the image


library/error-reporting.php
```diff
@@ -0,0 +1,36 @@
+<?php
+error_reporting(E_ALL | E_STRICT);
+ini_set('display_errors', true);
```
jamescrowley: not sure if this should actually be off?

library/error-reporting.php
```diff
@@ -0,0 +1,36 @@
...
+function bootstrap_error_handler($errno, $errstr, $errfile, $errline)
+{
+    switch ($errno) {
+    case E_WARNING:
+    case E_NOTICE:
+    case E_STRICT:
+        // we should really not be supressing these :(
+        // but maintaining backwards compatibility for now as
+        // there are loads of problems if you don't
+        break;
```
jamescrowley: Also seems to still be triggering certain notice warnings with Sentry but not others??

ajax.php
```diff
@@ -2,5 +2,4 @@
 	$login = true;
 	require_once('library/core.php');
 	
-	error_reporting(0);
```
HaGuesto: What was this for @barto74 ?

barto74: Error reporting to the std output in an Ajax call will never give any desired results, so just needs to be off.

HaGuesto:
> Consolidate error reporting so Sentry can report these, and we show a error landing page even for code failures early in the loading of the library stack
> 
> Lots of warnings are currently being supressed but can't easily resolve them,
> so decided to preserve that behaviour.
> 
> Notes:
> 
> * now forcing the whole app to run through a single entry point again
>   to simplify where we hook up error handling. Hence including the htaccess necessary
>   for this.
> * if you're running from docker you'll need to do `docker-compose build` to rebuild the image

Could you add some more instructions what to test exactly?
And how to set it up for @barto74?

ajax.php
HaGuesto: @jamescrowley should we readd it then?

library/core.php
```diff
@@ -1,13 +1,8 @@
...
+    throw new Exception("This app must now be running through the single entry point. Is your web server config directing all php traffic to gcloud-entry.php?");
 
 define('CORE',true);
-# start a session, if it's not already started from our caller (ie gcloud-entry)
-if (session_status() == PHP_SESSION_NONE) {
-    session_start();
-}
-
-# load configuration file
-require_once(__DIR__.'/config.php');
```
HaGuesto: @jamescrowley Where did you move the loading of the config file to?

PR #43, opened 2019-07-12
Unify error login notifications in php files and remove them from DB

DurkoMatko:
**change:** all login notifications are now defined in php files (just once)
**delete**: not needed minimalDb values for notification texts


library/core.php
```diff
@@ -52,5 +52,8 @@
 # functions that are app specific but need to available globally
 require_once('functions.php');
 
+define('GENERIC_LOGIN_ERROR',"This email either does not have an active account associated with it, or access has expired. Contact your coordinator to enable your account.");
```
HaGuesto: @DurkoMatko could you please put them in a seperate file in library lib and add `require_once`

DurkoMatko: Thought about it but wasn't sure if `/lib` isn't just for framework specific stuff. Done.

PR #47, opened 2019-07-12
Resort Stockroom and add firstline

HaGuesto:


DurkoMatko:
Hmm..I wanted to test this but my Stockroom is always empty :D I guess I'm doing something wrong...I set my Product to be always shown in Stockroom and I also created a box with 200 items of that product in location Stockroom. My stockroom is still empty tho..

HaGuesto:
> Hmm..I wanted to test this but my Stockroom is always empty :D I guess I'm doing something wrong...I set my Product to be always shown in Stockroom and I also created a box with 200 items of that product in location Stockroom. My stockroom is still empty tho..

@Martin did you tick the Stockroom checkbox in your stockroom location? Otherwise, it does not work. 
Maybe something to make a quick call about ?

image

PR #57, opened 2019-07-14
add: cypress added cicleci build (against staging env)

DurkoMatko:
Introduce Cypress browser tests on our staging environment. Note this won't actually stop a deploy to production right now, as that is just controlled by a merge to 'production' branch. Ideally we would run these tests on every PR and block merges that way - but requires more work in our set up first so we can run MySQL in our CircleCI.


jamescrowley:
I'm looking at what we need to do to get this working

jamescrowley:
This is now passing/failing in the same way my local tests are. Need to resolve those issues and then uncomment the config to only run on master builds (till we can get this working against individual containers for the PR)

PR #58, opened 2019-07-15
 disabled 'order from stockroom for invisible stockrooms and if item …

naphets123:
…is ordered already


include/stock.php
```diff
@@ -24,7 +24,9 @@
 
 		listsetting('manualquery',true);
 		
-		$data = getlistdata('SELECT stock.*, cu.naam AS ordered_name, cu2.naam AS picked_name, SUBSTRING(stock.comments,1, 25) AS shortcomment, g.label AS gender, p.name AS product, s.label AS size, l.label AS location, IF(DATEDIFF(now(),stock.modified) > 90,1,0) AS oldbox FROM '.$table.'
+		$data = getlistdata('SELECT stock.*, cu.naam AS ordered_name, cu2.naam AS picked_name, SUBSTRING(stock.comments,1, 25) AS shortcomment, g.label AS gender, p.name AS product, s.label AS size, l.label AS location, IF(DATEDIFF(now(),stock.modified) > 90,1,0) AS oldbox ,
+		IF(l.visible = 0 OR ordered !=0 OR l.label = "Stockroom",True,False) AS disableifistrue
```
HaGuesto: @naphets123 I'm quite sure that not all conditions will work.
Please test each condition.
It should look like this 

`IF(NOT l.visible OR stock.ordered OR stock.ordered IS NOT NULL OR l.container_stock,True,False)`

Please change the label condition. In the locations table there is a bool called container_stock which defines if a location is a stockroom. Sorry for the naming.

include/stock-list_edit.php
```diff
@@ -26,6 +25,7 @@
 				l.camp_id = '.$_SESSION['camp']['id'].' AS visible,
 				l.camp_id != '.$_SESSION['camp']['id'].' AS preventdelete,
 				l.camp_id != '.$_SESSION['camp']['id'].' AS preventedit,
+				IF(l.visible = 0 OR ordered !=0 OR l.label = "Stockroom",True,False) AS disableifistrue,
```
HaGuesto: See other comment.

PR #74, opened 2019-07-22
Feature reset tokens on cycle start setting

DurkoMatko:
**add:** setting and logic to reset tokens on new cycle
**add:** resettokens column in camps (migration)


db/migrations/20190721183024_add_resettoken_column_in_bases.php
```diff
@@ -0,0 +1,289 @@
...
+        $this->table('camps', [
+                'id' => false,
+                'primary_key' => ['id'],
+                'engine' => 'InnoDB',
+                'encoding' => 'utf8',
+                'collation' => 'utf8_general_ci',
+                'comment' => '',
+                'row_format' => 'DYNAMIC',
+            ])
+            ->addColumn('resettokens', 'string', [
```
HaGuesto: Line 19 until 25 are the only ones needed. The rest was created by the generate funtion. Please see the migrations before and do it like there and do not use the generate function.

Please check the phinx doc to see how to write simple migrations. It is actually super easy.

include/give2all.php
```diff
@@ -37,6 +37,11 @@
 
 				$currentdrops = db_value('SELECT SUM(drops) FROM transactions AS t WHERE people_id = :people_id',array('people_id'=>$person));
 				
+				if($_SESSION['camp']['resettokens']){
+					db_query('INSERT INTO transactions (people_id,description,drops,transaction_date,user_id) VALUES (:people_id,:description,:drops,NOW(),:user_id)',array('people_id'=>$person,'description'=>"Reset of remaining tokens before new cycle",'drops'=>(-1) * $currentdrops,'user_id'=>$_SESSION['user']['id']));
```
HaGuesto: Could u check with Roanna if the text is ok?

HaGuesto: I mean for the description.

DurkoMatko: It's just a DB value tho. I mean, I can check with her, but it's just a description of transaction. Users never see it or?

db/migrations/20190721183024_add_resettoken_column_in_bases.php
DurkoMatko: Done

include/give2all.php
DurkoMatko: Done

PR #75, opened 2019-07-25
Date signature

naphets123:
Solved as discussed before and stated in the comments, but right now the signature-date is reset if the signature is saved again, even if it did not change.


include/people_edit.php
```diff
@@ -81,7 +81,7 @@
 
 		#$formdata = $formbuttons = '';
 		$side['approvalsigned'] = db_value('SELECT approvalsigned FROM people WHERE id = :id', array('id'=>$id));
-
+		$side['date_of_signature'] = db_value('SELECT date_of_signature FROM people WHERE id = :id', array('id'=>$id));
```
jamescrowley: @naphets123 I realise there's a tonne of other inefficient SQL queries below this, but in this instance I think we could at least get both this and the approvalsigned info in a single SQL query given the WHERE is the same? ie

     SELECT approvalsigned, date_of_signature FROM people WHERE id = :id

using db_row instead of db_value?

naphets123: i moved them into a single query now, also moved the handling of both in custom.js into a single eventhandler

assets/js/custom.js
```diff
@@ -73,6 +73,15 @@ $(function() {
             );
         }
     });
+
+    $("#sig").signature({
```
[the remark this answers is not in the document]
naphets123: Ok, I didn't know about this, I hope it is correct this time

include/people_edit.php
jamescrowley: 🎉

assets/js/custom.js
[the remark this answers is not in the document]
jamescrowley: my bad - custom.js isn't actually included in the minified JS for some reason anyway 🤦‍♂

assets/js/custom.js
```diff
@@ -71,8 +71,14 @@ $(function() {
                 "checked",
                 !$("#sig").signature("isEmpty")
             );
+            if (!$("#sig").signature("isEmpty")) {$("#field_date_of_signature").val((new Date()).toISOString().slice(0, 19).replace('T', ' '));}
```
jamescrowley: I think this field is missing from the template somewhere?

include/people_edit.php
```diff
@@ -80,8 +80,7 @@
 		$ajaxaside = new Zmarty;
 
 		#$formdata = $formbuttons = '';
-		$side['approvalsigned'] = db_value('SELECT approvalsigned FROM people WHERE id = :id', array('id'=>$id));
-
+		$side['approvalsigned'] = $side['date_of_signature'] = db_row('SELECT approvalsigned, date_of_signature FROM people WHERE id = :id', array('id'=>$id));
```
jamescrowley: Does this work for you? I'm not sure this syntax does what you think it does... :grin:

naphets123: I changed it, but it seemed to work

assets/js/custom.js
naphets123: What do you mean by that?

include/people_edit.php
naphets123: It actually did, but i agree that it feels very wrong, I changed it

jamescrowley:
To record the discussion previously, at the moment this isn't visible in the UI, and updates every time someone saves the page even if it doesn't change. We'll need to look at this further before we can merge.

HaGuesto:
Trello card 
-------------
https://trello.com/c/AsWS8MEq

PR #77, opened 2019-07-26
minimal solution in template, needs correct IHA-organisation-label

naphets123:
This is the minimal solution, without a database, it still requires the correct organization-label(currently just "IHA")

### DIFF
--- a/templates/cms_form_signature.tpl
+++ b/templates/cms_form_signature.tpl
@@ -37,26 +37,26 @@
 			</ul>
 			<p>You do on your own provide us with your personal information in conversation with representatives from {$currentOrg['label']}. We need this information to ensure that you are part of our distribution and/or other services provided by the organisation, and to make sure that we have enough equipment for our beneficiaries.</p>
 			<p>{$currentOrg['label']} does not share your personal information with other parties.</p>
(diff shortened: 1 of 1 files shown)


templates/cms_form_signature.tpl
```diff
@@ -22,7 +37,7 @@
 			</ul>
 			<p>You do on your own provide us with your personal information in conversation with representatives from {$currentOrg['label']}. We need this information to ensure that you are part of our distribution and/or other services provided by the organisation, and to make sure that we have enough equipment for our beneficiaries.</p>
 			<p>{$currentOrg['label']} does not share your personal information with other parties.</p>
-			<p>To access the mentioned services provided by {$currentOrg['label']}, you must agree that we can continue to process and retain this information about you.</p>
+			<p>{$privacytext}To access the mentioned services provided by {$currentOrg['label']}, you must agree that we can continue to process and retain this information about you.</p>
```
jamescrowley: What was {$privacytext} for? I don't think it's defined anywhere?

naphets123: no this is a mistake, this shouldn't be there anymore

jamescrowley: 👍 I'll remove

PR #78, opened 2019-07-27
Usergroup bases display

naphets123:
The base names are visible, and the base column is hidden if the current user has access to less than 2 bases and is not an admin. The admin condition is there to accommodate some_admin, which has no usergroup, I am not sure if there is a backdoor like this in the running implementation, if no, we possibly could remove the condition on admin, since normal users don't see the usergroups menu anyway

### DIFF
--- a/include/cms_usergroups.php
+++ b/include/cms_usergroups.php
@@ -31,20 +31,23 @@
 				'.(!$_SESSION['user']['is_admin']?' AND l.level < '.intval($_SESSION['usergroup']['userlevel']):'').'
 			GROUP BY g.id');
 			
-		$num_camps=db_value('
-		SELECT 
-			COUNT(c.name)
-		FROM cms_usergroups_camps AS x
-		INNER JOIN cms_usergroups AS g ON x.cms_usergroups_id = g.id
-		INNER JOIN cms_users AS us ON (us.cms_usergroups_id = g.id)
-		INNER JOIN camps as c ON c.id = x.camp_id 
-		WHERE us.id = '.$_SESSION['user']['id'].' ');
-		
 
 		addcolumn('text','Name','label');
 		addcolumn('text','Level','userlevel');
-		if ($num_camps>1 OR $_SESSION['user']['is_admin']){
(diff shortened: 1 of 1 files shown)


include/cms_usergroups.php
```diff
@@ -30,11 +30,22 @@
 			WHERE (NOT g.deleted OR g.deleted IS NULL) AND g.organisation_id = '.$_SESSION['organisation']['id'].'
 				'.(!$_SESSION['user']['is_admin']?' AND l.level < '.intval($_SESSION['usergroup']['userlevel']):'').'
 			GROUP BY g.id');
+			
+		$num_camps=db_value('
+		SELECT 
+			COUNT(c.name)
+		FROM cms_usergroups_camps AS x
+		INNER JOIN cms_usergroups AS g ON x.cms_usergroups_id = g.id
+		INNER JOIN cms_users AS us ON (us.cms_usergroups_id = g.id) OR us.id = 1
```
jamescrowley: Why do we need 'OR us.id = 1' ?

jamescrowley:
Thanks @naphets123 !

include/cms_usergroups.php
naphets123: ok the only reason for this one is that the default-account(some admin) has neither an usergroup_id nor an organisation_id and i didn't want it to be invisible for that one but I see why it is undesirable to have this exception here

aerinsol:
We are showing admin users in the management consoles but only to people of
that same level or higher

include/cms_usergroups.php
jamescrowley: does that mean a boxwise admin user (ie is_admin=1) will never have those? or specifically userid 1? I couldn't see any other exceptions in the code looking for user id 1 so that's why I'm wondering :)

naphets123: The answer about this depends on the policy in setting up accounts/usergroups. I think we could change it to is_admin=1. I just realized i got confused because by "Test R4 admin" which has usergroups and organisation. I failed to notice that this account is not an admin, and therefore thought "some admin" is a special case.

PR #80, opened 2019-07-28
Save and create

naphets123:
This enters a confirmation in case of save and new is chosen in box menu


jamescrowley:
@naphets123 looking good! Just noticed a bit of a layout issue when the browser is slightly smaller than normal: 

Screenshot 2019-08-02 at 14 07 30

if you get time do you think you could have a check of the CSS to see if we can fix?

naphets123:
It might be easier for now to just remove the "confirmation" title, which would make it more consistent with the other option("save and close"). The title is set via the addfield method and i don't know which .css file is attached to this.

PR #84, opened 2019-08-03
Shopping cart in checkout

DurkoMatko:
**add:** shopping cart in checkout ("Add to cart" button, change quantity or delete option for items in cart, whole shopping cart sum) instead of submitting transactions product by product
**change:** changing selected product doesn't do backend request to get price of the product, prices are loaded during the first page load.


DurkoMatko:
@HaGuesto short call before merging this to master would be ideal. Have one remark/question to it.

DurkoMatko:
@HaGuesto 
Regarding point 2 - So if I delete both new costs fields and change the old one to the value of the shopping cart, there won't be any visible field which shows how much are you about to add to the cart. Is that ok? Cuz the old value will be now showing the value of the whole cart, but there's nothing showing value of what's currently selected in the form.

HaGuesto:
I see what you mean.
I think it is fine. 
The cart value is the most important data here. It is the value which is transfered when the save button is clicked.

The value of each item of the shopping cart is shown in the list of the cart and that is enough. Anything more I feel is confusing. 


Additional comment:
I also think that the warning of "not having enough tokens" - should only appear if the cart value is greater than the tokens they have.

DurkoMatko:
fixed points 2 and 3

HaGuesto:
https://trello.com/c/IUMhudab

This feature is not finished, but good enough to be merged.

PR #95, opened 2019-08-07
Easy box history

naphets123:


templates/cms_form_created.tpl
```diff
@@ -5,6 +5,7 @@
 {if $data['modified']}
 	{$translate['cms_form_modifiedby']} {if $data['modified_by']}{$data['modified_by']}{else}<em>{$translate['cms_form_createdunknown']}</em>{/if}<br />{$translate['cms_form_ondate']} {$data['modified']} 
 {/if}
-
+	{if !({$smarty.get.origin} == 'stock')}
```
jamescrowley: @naphets123 it would be helpful to have some more context in the PR description on *why* here too - not clear why if origin is not stock then we want the history? Thanks!

naphets123: To keep the behavior of other parts independent, since we only wanted to change the display of box history

aerinsol: @naphets123 Can you confirm that you are able to directly add comments in the code for context, as discussed in our sprint plan?

PR #96, opened 2019-08-08
Readable box history

naphets123:


library/lib/tools.php
```diff
@@ -32,20 +32,21 @@ function showHistory($table,$id) {
...
-			}
-			$row['changes'] .= '; ';
+		if ($row['changes'] == 'items') {$row['changes'] = "changed the number of items from ".$row['from_int']." to ".$row['to_int'];}
+		else if ($row['changes'] == 'location_id') {
+			$loc_ids = array($row['from_int'],$row['to_int']);
+			$loc_orig = db_row('SELECT locations.label FROM locations WHERE locations.id = :id_orig',array('id_orig'=>$loc_ids[0]));
+			$loc_dest = db_row('SELECT locations.label FROM locations WHERE locations.id = :id_dest',array('id_dest'=>$loc_ids[1]));
+			$row['changes'] = "changed box location from ".$loc_orig['label']." to ".$loc_dest['label'];}
+		else if ($row['changes'] == 'Record created'){
+			$row['changes'] = " created the box";}
```
jamescrowley: Doesn't this break the output for non-box pages, given this is a generic showHistory function?

jamescrowley:
@naphets123 it would be helpful to have some more context in the PR description on *why* you took the approach you did. for instance, why are we showing the 'naam' field when the table is 'stock' ?

PR #98, opened 2019-08-10
Auto sum boxes

naphets123:


jamescrowley:
@naphets123 is there anything you think we can do here to improve the code generally in this area? By the time there's listfilter1,2,3 & 4 with similar code, that starts to feel like we could try and consolidate?

naphets123:
Sure, especially since thery are the same function, but in the past we stuck to not rewriting things, so did it this way. I wanted to discuss this later today

jamescrowley:
Trello card: https://trello.com/c/07ec5P3K/252-in-boxes-auto-sum-the-number-of-boxes-items-for-displayed-boxes-medium

PR #100, opened 2019-08-11
Enable "Add and new" button in beneficiary menu

HaGuesto:
The reason for this bug needs more investigation.

@aerinsol @jamescrowley The helper function `addformbutton` does not work for some reason anymore. It does not add entries to the global array `$formbuttons` anymore. This affects any instance of addformbutton throughout the code. E.g., the "Submit and edit" button in the Boxes menu does not work anymore, too.

 Were there any changes recently which could just this helper function?

#### Related trello card
https://trello.com/c/AJFbDch4

### DIFF
--- a/include/people_edit.php
+++ b/include/people_edit.php
@@ -236,8 +236,10 @@
 	addfield('created','Created','created',array('aside'=>true));
 
 	if ($id) addformbutton('submitandedit',$translate['cms_form_save']);
-	else addformbutton('submitandnew',"Save and new");
-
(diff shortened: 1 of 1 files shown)


jamescrowley:
@HaGuesto pretty sure it's tracing function in the gcloud-entry.php file.  I had to declare all global variables there, and looks like I've missed $formbuttons fro the list.

HaGuesto:
@jamescrowley I closed the PR because it does not make sense to do this work-around if you know the issue.

PR #105, opened 2019-08-12
Enable remote debugging support for the docker containers

jamescrowley:
While trying to profile to diagnose performance issues, I needed to enable Xdebug to get the profiling information out. It also enables line-by-line step through/debugging/breakpoints so pushing here as may be helpful to diagnose future issues.

Enable remote debugging support for the docker containers  …
* The dockerfile was updated to include the xdebug extension, and be automatically enabled in the php.ini
* Included a VS code configuration to support XDebugging, but you'll need to install the PHP Debug extension
* The XDEBUG_CONFIG environment variable configures XDebug. The remote host alias allows XDebug to connect to your debugging client (VS code or similar). It needs the 'host.docker.internal' to access the host rather than the container. Note this variable may differ on Linux, but works on MacOS and Windows.

### DIFF
--- a/.docker/php/Dockerfile
+++ b/.docker/php/Dockerfile
@@ -3,3 +3,5 @@ FROM php:7.2.2-apache
 RUN a2enmod rewrite
 # PHP extensions
 RUN docker-php-ext-install pdo_mysql exif
+# Enable xdebug for debugging & profiling
+RUN pecl install xdebug && docker-php-ext-enable xdebug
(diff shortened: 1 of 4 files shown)


HaGuesto:
I'm starting to hate xdebug.

I only managed that it catches my errors, but not my breakpoints.

The host `docker.host.internal` does not resolve to an ip on linux, because it is not specified. I have not found a way to specify the remote address properly yet.

https://github.com/docker/for-linux/issues/264

HaGuesto:
Added a commit and had to move it to a different pull request #131

PR #109, opened 2019-08-12
Replace any hard-coded HTTP urls to HTTPS

jamescrowley:
On the mobile login page, Cypress is causing errors when resources are loaded via HTTPS but then there's a HTTP redirect. Figured I might as well replace all occurrences rather than this one specific. Part of the work for https://github.com/boxwise/dropapp/pull/57


aerinsol:
Nice!!

PR #125, opened 2019-08-19
adjust stockroom data query

HaGuesto:
Things changed:
- show menu only if a stockroom location is defined. Otherwise, throw an error.
- only show products of their own camps
- merge two separate queries and add a having clause

Trello card
https://trello.com/c/ADBUaDoN


HaGuesto:
@jamescrowley I'm going to self merge it because it is an important one for IHA.

PR #130, opened 2019-08-25
Escape all data passed to smarty templates by default

jamescrowley:
There are currently lots of potential XSS vulnerabilities - the latest I noticed was message strings being passed in the URL and served up as-is.

 Switching this flag on will mean Smarty will encode all provided input to make it 'safe' for HTML output by default. 

We will need to use 'nofilter' on anywhere that is 'correctly' unescaped - so this PR may break some screens. In general, best practice means any HTML generation should be happening in the template, so fingers crossed there's not too many places this happens.

### DIFF
--- a/build/compile-smarty.php
+++ b/build/compile-smarty.php
@@ -14,7 +14,6 @@ $deploymentRootFolder = '/srv';
 # The hash was reverse engineered from smarty_internal_resource_file.php
 $smarty = new Zmarty;
 precompileTemplates($smarty);
-echo "Re-mapping templates from '$originalTemplateDir' to '$deployedTemplateDir'\n";
 rewriteAllTemplateHashes($smarty, $deploymentRootFolder);
 
 function precompileTemplates($smarty) {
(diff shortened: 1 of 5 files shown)


jamescrowley:
Other likely problems:

https://github.com/boxwise/dropapp/blob/master/include/stock-list_edit.php#L15
https://github.com/boxwise/dropapp/blob/master/library/ajax/laundry.php#L20

HaGuesto:
> https://github.com/boxwise/dropapp/blob/master/include/stock-list_edit.php#L15
Titles are included [directly in `cms_form.tpl`](https://github.com/boxwise/dropapp/blob/d89ab35ba1a5962e66da61448d99f948582e1814/templates/cms_form.tpl#L4) 

> https://github.com/boxwise/dropapp/blob/master/library/ajax/laundry.php#L20
This one is a horrible hack [which uses `cms_form_custom.tpl`](
https://github.com/boxwise/dropapp/blob/943049a0ab74230f72a197a4e262ace6eec1bbdd/templates/cms_form_custom.tpl#L4)

PR #92, opened 2019-08-04
Enable warnings in local development mode

jamescrowley:
This PR is enabling PHP warnings for local development, so we can slowly get the code base to a point there are no warnings. Having the warnings enabled should save us some occasional 'gotchas' with variable scopes and declarations - already found one bug whilst creating this PR having turned on the warnings

PR #132, opened 2019-08-25
Fix for exception caught with Xdebug

HaGuesto:


jamescrowley:
Ah 👍 thought you were talking about a different exception when you mentioned the other day. I fixed this one in https://github.com/boxwise/dropapp/pull/92 too as the warnings picked it up, but I've removed from there so we can just merge this instead.

PR #134, opened 2019-08-25
Validate qr_id when box is created on mobile

HaGuesto:
This is a minimal solution. I just validated with an if statement if `$_POST['qr_id']` is set and not zero when the box form is submitted on mobile phones. 

Trello card
--------
https://trello.com/c/nHdAMVbi

### DIFF
--- a/mobile/savebox.php
+++ b/mobile/savebox.php
@@ -1,5 +1,6 @@
 <?php
 
+if (isset($_POST)) {
 	if(!intval($_POST['box_id'])) {
 		$new = true;
 		do {
@@ -37,3 +38,6 @@
 	$data['barcode'] = db_value('SELECT code FROM qr WHERE id = :id',array('id'=>$box['qr_id']));
 
 	redirect('?barcode='.$data['barcode'].'&message='.$message);	
+} else {
+	redirect('?warning=1&message=Something went wrong! Please try again!');
+}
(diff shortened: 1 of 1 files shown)


jamescrowley:
@HaGuesto it was this bit in the existing code that didn't make sense to me?

         if(!intval($_POST['box_id'])) {
		$new = true;
		do {
			$_POST['box_id'] = generateBoxID();
			} while(db_value('SELECT COUNT(id) FROM stock WHERE box_id = :box_id',array('box_id'=>$_POST['box_id'])));
		}

do you know what that's doing? I'd have thought if there was no POST box_id then we should stop then and there, no?

HaGuesto:
> @HaGuesto it was this bit in the existing code that didn't make sense to me?
> 
> ```
>      if(!intval($_POST['box_id'])) {
> 	$new = true;
> 	do {
> 		$_POST['box_id'] = generateBoxID();
> 		} while(db_value('SELECT COUNT(id) FROM stock WHERE box_id = :box_id',array('box_id'=>$_POST['box_id'])));
> 	}
> ```
> 
> do you know what that's doing? I'd have thought if there was no POST box_id then we should stop then and there, no?

`box_id` is the 6-digit unique number for the box which is different to the `qr_id` which is the id of the qr-hash in the `qr`-db-table.

If a qr-code is scanned which has no box connected to it (a new fresh qr-code) then this if clause is true and generates a random `box_id`.

HaGuesto:
@jamescrowley

HaGuesto:
@jamescrowley updated

PR #138, opened 2019-08-25
Introduce auto formatting of PHP code

jamescrowley:
So we don't have to think...

CircleCI will now *fail* if there is any code requiring listing fixes. 

This is done on the assumption that the vscode-php-cs-fixer plugin is running in VSCode and applying the auto format on save.

Alternatively, we can run

     php vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix . --rules @PhpCsFixer

in the root manually.


aerinsol:
Woohoo!!

PR #139, opened 2019-08-25
Enable id-card printing

HaGuesto:


pdf/bicyclecard.php
```diff
@@ -36,7 +36,7 @@
 		$settings['upload_dir'].'/people/'.intval($p['id']).'.jpg':'');
 		
 	
-	if($p['bicycletraining']) {
+	// if($p['bicycletraining']) {
```
jamescrowley: yo @HaGuesto, just a gentle nudge - what's your thinking re competing out vs just removing? Unless it really is a temporary change, I'd always remove it entirely. That’s what git is for :grin:

PR #149, opened 2019-08-27
Enable warnings in local development mode

jamescrowley:
Reverts #147 to reinstate #92

This PR is enabling PHP warnings for local development, so we can slowly get the code base to a point there are no warnings. Having the warnings enabled should save us some occasional 'gotchas' with variable scopes and declarations - already found one bug whilst creating this PR having turned on the warnings

### DIFF
--- a/index.php
+++ b/index.php
@@ -1,12 +1,15 @@
 <?php
-	
+	$login = false;
+	$ajax = false;
+	$mobile = false;
+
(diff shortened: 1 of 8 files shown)


jamescrowley:
@HaGuesto heads up I changed this so it only logs to the stderr (console) rather than breaking the UI with warnings - there were too many places where redirects no longer worked.

PR #159, opened 2019-09-05
Correciton box history

naphets123:
Removed two cases where complete messages are inserted into history table and changed to default form. Added case handling for all change-messages I could find for the box history.


naphets123:
https://trello.com/c/7RNuUJRQ/155-use-human-readable-names-for-box-history-instead-of-ids-tiny

HaGuesto:
@naphets123 Thanks for the PR. `showHistory` became a crazy function. I'm very happy that you did that.

PR #160, opened 2019-09-05
add a first login function and test 3.1.1

HaGuesto:


cypress/support/session.js
```diff
@@ -0,0 +1,39 @@
+// Standard Login method for all tests
+Cypress.Commands.add("LoginAjax", (userMail, userPassword, autologin) => {
+    cy.visit('/'); //not sure if needed. Somehow it does not work without it.
```
HaGuesto: @jamescrowley This is the wicked bit. I would like to get rid off this line, but it does not work otherwise.

jamescrowley: Pushed two commits that resolve this - need to use cy.request :) rather than making the AJAX request via jQuery...

jamescrowley:
@HaGuesto did you mean to closed this? I'd pushed the commits which fixed this issue on the branch, but they haven't been merged. And I don't think your tests have either?

jamescrowley:
Ah, I'll just close and add my commits again in a separate PR - but let me know if you need your tests in too

PR #163, opened 2019-09-06
Remove purchase_quick_edit which became obsolete because of shopping cart

HaGuesto:
I found in `magic.js` a function listening to the class `.change-count`.
It was a function which was used before in the checkout and should adjust the transactions if the amount was changed.

For now I just uncommented it to test if it is not used anymore.


jamescrowley:
@HaGuesto I'd be tempted to just remove everything outright. If it works, then we don't have to go back and remove the commented code. and if it doesn't, we just revert this commit?

jamescrowley:
Plus, is there any way we can get confidence in this before we merge? :D

HaGuesto:
> Plus, is there any way we can get confidence in this before we merge? :D

lol. I'm quite sure, like 95%, but the recent weeks have made me a bit more careful.

I'll update it.

PR #168, opened 2019-09-07
Disable "Add to cart" button for negative number of items

DurkoMatko:
**fix:** add to cart button disabled for negative number of items


jamescrowley:
@DurkoMatko Thanks! I noticed in my testing using 'onchange' on the number field means that the 'Add to cart' button doesn't re-enable until I click outside of the number field, which may be confusing. 

Did you happen to look at using 'min' on the field instead of custom JS to enforce this?

DurkoMatko:
@jamescrowley nope, I haven't. Will look at it today in the evening.

DurkoMatko:
@jamescrowley hmm 'min' doesn't work...there's also try to use it in generating QR codes. That field also is supposed to have min value (see qr.php) but the form.php which is responsible for generating fields seems to ignore it..No idea why.

I could add some handler on the field which would automatically revert value to 1 if user input is less then or equal to zero? Or spend more time trying to make the 'min' happen..but probably during the weekend first.

Not sure why you suggest listening to click event.

jamescrowley:
> @jamescrowley hmm 'min' doesn't work...there's also try to use it in generating QR codes. That field also is supposed to have min value (see qr.php) but the form.php which is responsible for generating fields seems to ignore it..No idea why.

Had a bit of a dig, pretty sure we need to add

      {if $element['min']}minlength="{$element['min']}"{/if}

to https://github.com/boxwise/dropapp/blob/master/templates/cms_form_number.tpl

> Not sure why you suggest listening to click event.

Sorry, ignore - I remember some cases you had to listen to click in addition to on change when the focus left fields... suspect it would have to be something else in this case, but if ^^^ works we shouldn't need it

DurkoMatko:
??? 
@jamescrowley No idea what happened here. I've rebased the branch to totally remove my initial commit with JS changes without leaving it in history. Then force-pushed. That automatically closed the PR. Then I've followed this article (https://evilmartians.com/chronicles/git-push---force-and-how-to-deal-with-it) to reopen the PR...and now it says it got successfully merged? Is it saying it got merged cuz there's no difference between this branch and master (because I've rebase removed the only commit which was different?) No idea...I'll just push the 'min' change from new branch

jamescrowley:
> @jamescrowley No idea what happened here. I've rebased the branch to totally remove my initial commit with JS changes without leaving it in history. Then force-pushed. That automatically closed the PR. 
Yeah, that happened because you ended up with no commits with changes after your re-base - not quite sure how that happened though. 

Now that you got it working on a new branch, you could do something like

```git checkout fix-negative-number-of-items-in-checkout
git reset add-min-attribute-to-number-input-template --hard
git push --force
```

which would check out this branch, reset it to the same as add-min-attribute-to-number-input-template, and push force here. But no need - I'll just review on the other PR

PR #172, opened 2019-09-08
Feature e2e cypress checkout tests

DurkoMatko:
**add:** cypress e2e checkout tests 

(2 tests specified in https://docs.google.com/spreadsheets/d/1sDhSsaVwNxAhGn1VYACDPVepZpvlwTkgy9nmeTaRy2Q/edit#gid=0 don't work yet but maybe as an initial PR, this can be merged)


assets/js/shoppingCart.js
```diff
@@ -158,10 +158,10 @@ $(document).ready(function() {
             $('#shopping_cart')
               .append("<tr><td>"
                   + item.name +'</td><td>'
```
jamescrowley: I realise not the focus of this PR, but just a heads up that whenever doing this kind of HTML construction, in order to be safe from XSS vulnerabilities, it's generally advisable to use the 'safe' DOM construction methods like createTextNode/textContent. Never trust user input :)

cypress/integration/5_1_CheckoutTests.js
```diff
@@ -0,0 +1,350 @@
...
+        testUser = $result.testUser;
+        testPwd = $result.testPwd;
+      });
+  });
+
+  beforeEach(function(){
+  });
+
+  function navigateToCheckout(){
+      cy.LoginAjax(testAdmin, testPwd, true);
```
jamescrowley: @DurkoMatko regarding your comment in Slack, if we have to pick one for now, I'd go with co-ordinator or volunteer for now. Admin users are only us, and we're the least important folk if something breaks :)

jamescrowley:
@DurkoMatko you've clearly had to jump through a lot of browser hoops to interact with the checkout... thank you for all the comments in the code! :grin:

Are there things that if we changed in the UI HTML that would make these tests much easier to write / understand? I'm thinking it might be hard for others to pick these up right now (me, at least!). I'm guessing some of the data-testid stuff might have been hard to flow down to the right places?

DurkoMatko:
@jamescrowley Not sure what could be done differently there...it's a mix of shopping cart elements with `data-testid` attributes (which were actually easy to add as I was familiar with checkout cuz of implementing the cart)...but the DOM elements generated by Smarty were quite a pain to work with...There are lot of various hidden elements which were being yielded by `cy.get()` instead of the once I was expected...so I had to specify more detailed DOM paths to get the correct elements etc etc...but not sure if something can be done better there.

cypress/integration/5_1_CheckoutTests.js
DurkoMatko: Done. Will push tomorrow along with `createTextNode/textContent`..

assets/js/shoppingCart.js
DurkoMatko: Nice, good to know.
I've almost fixed it now..the only issue is that I'm able to add products just to the table header which causes styling to be a bit off...I'm using `tableRef.insertRow()` and after that ` createTextNode/textContent.`... even if I add tbody, then rows are still added to header...dunno, maybe I'm just tired, will try tomorrow again...if not, I'd add missing padding for `td` tags in `thead` and it'll be all in header :D

cypress/integration/5_1_CheckoutTests.js
jamescrowley: @DurkoMatko if you make changes for createTextNode etc, would you mind doing that in a separate PR? Will make reviewing easier, as it's an unrelated change :)

cypress/integration/5_1_CheckoutTests.js
```diff
@@ -0,0 +1,350 @@
...
+    selectRandomFamily();
+    selectProductWorthZero();
+    let randomCount = Math.floor(Math.random() * 10 - 1) + 1;
+    typeProductQuantity(randomCount);
+    // add to cart should be enabled even if product has 0 tokens value
+    cy.get("button[data-testid='add-to-cart-button']").should('not.be.disabled');
+    clickAddToCartButton();
+    clickCheckoutSubmitButton();
+    // visible notification
+    cy.NotificationWithTextIsVisible("Shopping cart successfully submitted!");
```
jamescrowley: I like the readability of these tests with the descriptive functions @DurkoMatko ! Maybe there are other ways we can wrap these up? The more descriptive functions we have, the less comments describing what is going on we'd need.

cypress/integration/5_1_CheckoutTests.js
```diff
@@ -0,0 +1,350 @@
...
+      cy.get("body").then($body => {
+          expect($body.find("select[data-placeholder='Please select']").length).to.equal(2);
+      });
+  });
+
+  it('Select family in dropdown', () => {
+    navigateToCheckout();
+    cy.get("div[id='s2id_field_people_id']").click();
+    // get name of first beneficiary before clicking it
+    cy.get("ul[class='select2-results'] li").first().find("div").invoke('text').then((text) => {
```
jamescrowley: these are all select2 drop down related, right? perhaps these can be generalised, as there seems to be quite a few usages? (within selectRandomProduct, selectRandomFamily, here, 'Select product in dropdown', etc)

cypress/integration/5_1_CheckoutTests.js
```diff
@@ -0,0 +1,350 @@
...
+    })
+    .then(currentCartValue => {
+      expect(parseInt(currentCartValue)).to.equal(expectedCartPrice);
+    });
+  });
+
+  it('Cart value - decrementing product count', () => {
+    navigateToCheckout();
+    randomizeCartContent();
+    let increasedPriceBy;
```
jamescrowley: Wondering whether several of these tests would be significantly simpler if we just target specific products with data we already know about (and therefore don't need to select a random product quantity and random family etc?). I'm in two minds on this though, because I like that this is super-flexible regardless of the test data. Thoughts?

DurkoMatko: Hmm simpler in what sense? To write and read - for sure. Dunno, maybe we should make some "official" rules when it comes to testing and to what extent should it be randomized? If it was me, I'd probably keep this one in there like it is as I've already created that cart randomizer function so it'd be a waste to jut delete it now :D Plus it makes up for a more robust test for sure. But if talking generally, I might agree on writing tests with some hardcoded values from now on. It'd for sure easier and less time consuming. But then again, I assume these randomized tests will probably catch bit more potential issues. It's a tradeoff or? But if we decide to test with hardcoded values we know are in seed data, I wouldn't argue about it.

cypress/integration/5_1_CheckoutTests.js
DurkoMatko: Are those comments bad in general? I've just felt "motivated" so I was adding them there to make life of future contributors easier...but if it destroys readability, I can delete them.

cypress/integration/5_1_CheckoutTests.js
DurkoMatko: https://github.com/boxwise/dropapp/pull/174

cypress/integration/5_1_CheckoutTests.js
DurkoMatko: I've added several additional helper methods now. None of the `select2` related selectors are being used in test bodies, they are always used just withing helper with explanatory name. Exception is when they're used in .then block on the `$body`. It happens maybe 2 times..I've thought passing `$body` to extra helper function would be overkill.

DurkoMatko:
@jamescrowley no idea about how to make selectors more robust...those DOM elements are all generated by smarty...means they're templates...so I can't add any `data-testid` attributes there

cypress/integration/5_1_CheckoutTests.js
jamescrowley: Definitely needed if the code itself isn't self descriptive as to *what* (which cy.get('random complicated selector).stuff.click() doesn't). In an ideal world

1. Have self-describing code that makes it easy to understand what it does, so you don't need a comment explaining that
2. Comments explain the 'why' (which you also have in several cases, which is great). 

In this specific case

    // visible notification
    cy.NotificationWithTextIsVisible

isn't really needed as the function is descriptive.

There are probably others we could work on (1) which would mean a comment isn't needed - but that's around my original question about if we can make the whole field selection stuff much simpler (in the way it is when we use data-testid)

DurkoMatko:
@jamescrowley apart from the 2 outcommented unfinished tests, I feel this is finished now (for now)

jamescrowley:
@DurkoMatko yeah, I think you've created your own commit with those changes rather than an actual merge commit or rebase. But this is looking good. Once I've fixed master deployment (due to FK issues) I'll rebase and merge this I think. Then you can update with the final tests when you get a chance.

jamescrowley:
@DurkoMatko heads up I've force-pushed, and :
- updated to use loginAsVolunteer
- moved navigateToCheckout to a beforeEach call (rather than being explicitly called each time)
- skipped going home and clicking checkout and instead hard-coded the checkout url. I'm suggesting we take this approach in general - no need to check the navigation works. One hard coded destination url isn't a big deal if it does ever change, and saves lots of extra page loads during testing.
- kept the go home and click checkout test to test ^^ path

DurkoMatko:
@jamescrowley Thanks! You're right the merge was my own commit as I've added beforeEach hook and loginAsVolunteer there. But git should kinda recognize those changes from merge part were already in master before, or? And even more no idea why loginAsVolunteer and beforeEach weren't part of the final code and you had to add them again...I remember even mentioning them to the merge commit message...

PR #173, opened 2019-09-10
5 2 add beneficiary

naphets123:
Test for Add beneficiary: 5_2_3 changed from specified behaviour due to different behaviour from the app, combined 5_2_4 and 5_2_5 to single test

https://trello.com/c/BGS2MaYC/297-cypress-automated-testing-of-add-beneficiary


cypress/integration/5_2_Add_Beneficiary.js
```diff
@@ -0,0 +1,109 @@
...
+    before(function() {
+        cy.getCoordinatorUser().then(($result) => {
+            Test_user = $result.testCoordinator;
+            Test_passwd = $result.testPwd;
+          });
+
+    })
+
+    beforeEach(function() {
+        cy.LoginAjax(Test_user,Test_passwd,true);
```
jamescrowley: Given the common usages of this, I'm thinking we maybe should just add helpers

     cy.LoginAsCoordinator() 

etc? Rather than having to do cy.getCoordinatorUser() and so on? @naphets123 @DurkoMatko thoughts?

jamescrowley:
@naphets123 @DurkoMatko might be good for you both to take a look at each other's PRs too to see if there's anything you've learnt that should be applied in each others code? (not got a specific case in mind here, just a random thought)

cypress/integration/5_2_Add_Beneficiary.js
DurkoMatko: Hmm that's actually true.. Although if we manage to come up with that multi-user-level testing approach (some "mighty" function) where you write test and specify user levels you want to run it for, these helpers might not be needed at all. Dunno, maybe merge this now and iterate on logging-in approach next weekend? Cuz those changes will be needed in my PR as well, might be easier if we just merge both PRs (once ready) and have one extra PR to add either that mighty function or at least login helpers. Or? Just thoughts :D

jamescrowley: OK, happy to merge if you're both ok to co-ordinate on that next weekend :)

cypress/integration/5_2_Add_Beneficiary.js
```diff
@@ -0,0 +1,108 @@
...
+    cy.get("div[class='select2-result-label']").first().click();
+
+    }
+
+    it("5_2_1 Fill form, Save and close",() => {
+        CheckEmptyForm();
+        //check all the forms 
+        FillForm(Test_firstname,Test_lastname,Test_case_id,Test_gender);
+        cy.get("button").contains("Save and close").click();
+        cy.wait(500);
```
jamescrowley: Something is going awry if we need to use .wait() - check out https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-Waiting and see if we can figure an alternative strategy? I'm guessing @DurkoMatko  might have already faced this.

cypress/integration/5_2_Add_Beneficiary.js
```diff
@@ -0,0 +1,85 @@
...
+
+    })
+
+
+    it("5_2_4 Save and new check if new person in familyhead-dropdown + check if empty",() => {
+        //check all the forms 
+        FillForm(Test_firstname,Test_lastname,Test_case_id);
+        cy.get("button").contains("Save and new").click();
+        cy.get("span[class='noty_text']").contains(Test_firstname+" "+Test_lastname + " was added").should("be.visible");
+        cy.get("span[class='noty_text']").contains(Test_case_id).should("be.visible");
```
DurkoMatko: Maybe add some helper methods in these tests again? Especially if there's the same selector (e.g `cy.get("span[class='noty_text']")`) used several times, maybe having a helper method like where it's used just once would be nicer. For this notification, it'd be something like:

```
function getNotificationByText(text){
   return cy.get("span[class='noty_text']").contains(text);
}
```

You can also maybe add this new method to `/cypress/support/notifications.js` file so we can use it for other tests as well.

cypress/integration/5_2_Add_Beneficiary.js
```diff
@@ -0,0 +1,117 @@
...
+        SaveAndProgress("Save and new");
+        cy.notyTextNotificationWithTextIsVisible(Test_firstname+" "+Test_lastname + " was added");
+        cy.notyTextNotificationWithTextIsVisible(Test_case_id);
+        // Check for the familyhead after adding it above
+        cy.selectOptionByText("parent_id",Test_case_id +" "+ Test_firstname);
+        FillForm(Test_firstname2,Test_lastname,"");
+        SaveAndProgress("Save and close");
+        cy.notificationWithTextIsVisible(Test_firstname2+" "+Test_lastname + " was added");
+        CheckAssociation(Test_firstname,Test_firstname2);
+        DeleteTested(Test_firstname);
```
jamescrowley: What happens if the test fails (and therefore doesn't execute this line)? Will the subsequent test pass, or will it now always fail because the test user wouldn't have been deleted? @naphets123 @DurkoMatko

HaGuesto:
@DurkoMatko Stephan told me you should review it again and approve it.

PR #174, opened 2019-09-10
change: safe DOM contruction of shopping cart

DurkoMatko:
**change:** safe DOM contruction of shopping cart


assets/js/shoppingCart.js
```diff
@@ -155,13 +155,61 @@ $(document).ready(function() {
...
+          // product name cell
+          let tr = document.createElement("tr");
+          let td = document.createElement("td");
+          let newText = document.createTextNode(item.name);
+          td.appendChild(newText);
+          tr.appendChild(td);
+  
+          // product quantity cell
+          td = document.createElement("td");
+          let input = document.createElement("input");
```
jamescrowley: Could make this *slightly* nicer by using jQuery's syntax
```
jQuery('<div/>', {
    id: 'some-id',
    "class": 'some-class',
    title: 'now this div has a title!'
}).appendTo('#mySelector');
```

DurkoMatko: You mean everywhere? Like, should I replace every `.appendChild()` with this jQuery syntax? Or just those 2 inputs (quantity and delete button)?

jamescrowley: It was less about the appendChild and more about the document.createElement :) I'd stay consistent with one approach or the other - but feel free to apply or ignore, was more a heads up in case you preferred it than anything else.

jamescrowley: Though if we do any *more* client side rendering we should really switch to https://github.com/janl/mustache.js or something at the very least

DurkoMatko:
@jamescrowley I'm finished here. Seems to work now.

assets/js/shoppingCart.js
```diff
@@ -155,13 +155,50 @@ $(document).ready(function() {
...
+            type: "number",
+            "class": "form-control valid changeQuantity",
+            'data-testid': "changeQuantity",
+            'step': "1",
+            'min': "1",
+            'productid': item.id,
+            'value': item.count,
+          }).appendTo(amountCell);
+
+          // there seems to be bug in jQuery - it doesn't add attributes to table cell DOMs using .appendTo syntax (therefore we have to add it later with .attr)
```
jamescrowley: @DurkoMatko This seems to be because you're passing the inner element text through in the <td> construction , rather than using the text() method - I'll push an update, as that also means there's still XSS possibilities there, as I don't think jQuery will escape it when you concatenate it directly.

PR #175, opened 2019-09-11
3 2 Box creation tests

naphets123:
…ink to qr-code generator


naphets123:
https://trello.com/c/00TUcrC9/294-cypress-automated-testing-for-box-creation-desktop

cypress.json
```diff
@@ -1,6 +1,7 @@
 {
   "projectId": "tip3z9",
-  "baseUrl": "https://staging.boxwise.co",
+  "baseUrl": "http://localhost:8100",
```
jamescrowley: Don't think you meant to commit that?

jamescrowley:
@naphets123 can you explain exactly why this is need chromeWebsecurity to be disabled, and which bit of code requires it? (I couldn't figure it out looking at the PR). Do we need to actually load the QR code URL or can we just look at it?

cypress/integration/3_2_Box_creation_desktop.js
```diff
@@ -0,0 +1,132 @@
...
+
+    })
+
+    function selectRandomProduct(){
+        //select random Product, modified this from Martins function
+        cy.get("div[id='s2id_field_product_id']").click();
+        cy.get("body").then($body => {
+        cy.get("ul[class='select2-results'] li").first().click();})
+        cy.get("div[id='s2id_field_product_id']").then((text) =>{
+              Test_product = text.text().replace('Product','').trim();
```
jamescrowley: @naphets123 as discussed, accessing variables like this doesn't work reliably in Cypress. 

1. using closures - which *probably* makes most sense within a single test, so long as you don't have to nest too deeply - https://docs.cypress.io/guides/core-concepts/variables-and-aliases.html#Closures
```cy.get("h2").then((message) => {
           var Test_id1 = message.text().split('ID').pop().split('(write')[0].trim();
        cy.get(Test_id1).click(); // --< inline within the block you can a access the variable
});
```

or 2. using aliases

```
cy.get(`td:contains(${firstProductName})`)
          .its("text")
          .as("productText");
cy.get("@productText").then((productText) => {
   // do something with productText
});
```

cypress/integration/3_2_Box_creation_desktop.js
```diff
@@ -0,0 +1,132 @@
...
+        cy.getCoordinatorUser().then(($result) => {
+            Test_user = $result.testCoordinator;
+            Test_passwd = $result.testPwd;
+            });
+
+    });
+    
+    beforeEach(function() {
+        cy.LoginAjax(Test_user,Test_passwd,true);
+        cy.visit('/');
```
jamescrowley: Rather than going home, and then navigating to the stock page, I'd suggest we navigate straight to where we need to go - hard coding the URL is fine. Should save an extra unnecessary page load, and while the URL may change, it wouldn't be a big deal to fix. If you wanted, perhaps could have one test which loads the home page and checks the URL matches the one we've hard coded, but :shrug:

cypress/integration/3_2_Box_creation_desktop.js
```diff
@@ -0,0 +1,132 @@
...
+        cy.get("a[class=menu_stock]").last().click();
+        cy.get("a.new-page.item-add").click();
```
jamescrowley: Do we need to do this action on every test? I was looking at the QR code one (to try and debug the chromeWebSecurity issue). As far as I can see, we then immediately go back to the stock page again?

cypress/integration/3_2_Box_creation_desktop.js
```diff
@@ -0,0 +1,87 @@
...
+        cy.loginAsVolunteer();
+        cy.visit('/?action=stock_edit&origin=stock');
+    })
+
+
+    function CheckBoxCreated(Id,productname,size,location,items) {
+        cy.get("a[class='menu_stock']").last().click()
+        cy.get("input[class ='form-control input-sm']").type(Id)
+        cy.get("span[class ='input-group-btn']").click();
+        cy.get("td[class = 'list-level-0 list-column-product").should('contain',productname)
```
DurkoMatko: Adding those helper human-readable functions for all tests would be cool to have as we talked about in slack. You can check them out it test files for e.g 5_1, 3_3 or 3_1 to see how I did it after James suggested it to me.

DurkoMatko:
Hmm I can't find the `data-testid`being used anywhere. They be added in `stock_edit.php` file in addField method. Adding something like `testid` => 'YOURVALUE' should work. It's already present in the repo in other files, just search for it. Then you cann access those DOM elements using that rather the id or class or so.

cypress/integration/3_2_Box_creation_desktop.js
```diff
@@ -0,0 +1,87 @@
...
+        cy.selectOptionByText("product_id", product)
+        cy.get("input[id='field_items']").click().type(items);  
+        cy.selectOptionByText("size_id",size)
+        cy.selectOptionByText("location_id",location)
+    }
+    
+    function CheckEmpty(){
+        cy.get("span[id='select2-chosen-1']").contains("Please select").should("be.visible")
+        cy.get("span[id='select2-chosen-2']").contains("Please select").should("be.visible")
+        cy.get("span[id='select2-chosen-3']").contains("Please select").should("be.visible")
```
DurkoMatko: To interact with select2 element nicely, there are some helper functions specified in `/cypress/support/select2-interaction-methods.js`. You can use them like this:

`cy.getSelectedValueInDropDown("people_id").contains("Please select").should('exist');`

cypress/integration/3_2_Box_creation_desktop.js
```diff
@@ -0,0 +1,87 @@
+context("Box_creation_tests", () => {
+    let Test_location = "TestWarehouse";
+    let Test_number = "100";
+    let Test_product = "Jeans Female";
+    let productname  = "Jeans"; //in the box creation product name is automatically concatenated with gender-info, therefore we use this variable to store a processed version for checks in the boxes menu
```
DurkoMatko: maybe unify the variable name to the same _Naming_convention_ you're using? (I personally wouldn't mind but others always bully me for this at work when I get reviewed :D so I guess it's a good practice)

jamescrowley:
All looks good, but same question as the other PR - if this test fails, will it then permanently fail until we fix the test data (as it won't have 'deleted' the box that it has created) ?

HaGuesto:
@jamescrowley @naphets123 I talked with Stephan yesterday about his PR and thought we should just delete all boxes that are not in the seed each time before we run the test to start of all tests with the same data.

I would propose a global cypress function to do that.
I checked @DurkoMatko mobile Box creation tests and the deletion of his created boxes is missing there.

naphets123:
I added a function that deletes all boxes if any exist.

PR #177, opened 2019-09-11
Reset password null

naphets123:


naphets123:
https://trello.com/c/fyVu0Dvq/275-cannot-reset-passwords-if-your-validfirstday-and-validlastday-are-set-to-null-rather-than-0000-00-00-tiny

library/ajax/reset.php
```diff
@@ -2,8 +2,8 @@
 
 $row = db_row('SELECT *, "org" AS usertype FROM cms_users WHERE email != "" 
 	AND email = :email 
-	AND ((valid_firstday < NOW() AND valid_lastday > NOW()) OR UNIX_TIMESTAMP(valid_lastday) = 0)
-	AND (NOT deleted OR deleted IS NULL)', ['email' => $_POST['email']]);
+	AND ((valid_firstday < NOW() AND valid_lastday > NOW()) OR UNIX_TIMESTAMP(valid_lastday) = 0 OR valid_lastday == NULL)
```
jamescrowley: Thanks @naphets123 . Do you have any idea how valid_lastday ended up being NULL in the first place? I'm assuming elsewhere it's set to the equivalent of 0, given the existing clause of `UNIX_TIMESTAMP(valid_lastday) = 0`

jamescrowley:
For expediency, will just accept this for now and worry about the data inconsistencies later, don't think it's worth the effort for the level of impact / ease of fixing if it's an issue. Thanks @naphets123 !

PR #181, opened 2019-09-13
Cypress login helpers

jamescrowley:
1. I've updated the AJAX-based sign in with my original fix that got lost somewhere
2. I cleaned up the use of the .env variables a bit so we don't have to deal with async to read them.
3. You now have:

* cy.loginAsVolunteer
* cy.loginAsAdmin
* cy.loginAsCoordinator

available as helpers. 

4. I made the remaining helpers camelCase to be consistent


aerinsol:
Yay James!

PR #184, opened 2019-09-21
Remove unused Db library code

jamescrowley:
https://trello.com/c/FT4mJYcL/261-manage-beneficiaries-performance 

While investigating the performance of the beneficiaries page - and specifically https://github.com/boxwise/dropapp/blob/5eea7160ef85e4a5d2ed68ef1450e0a0c111d16e/library/lib/list.php#L492-L495 and https://github.com/boxwise/dropapp/blob/5eea7160ef85e4a5d2ed68ef1450e0a0c111d16e/library/lib/list.php#L399-L408 , I have noticed that (a) only the cms_functions and people table have a parent_id (the problematic performance scenario and (b) that we have no functions ending in _content.

As such, can simplify a bunch of code. This is the first step.

We have no tables that end in _content, so none of these code paths are used. Identified while trying to optimise getlistdata. To do this PR, I
- searched for all occurrences of _content in our code base
- removed the direct code using it
- removed any dependent code using it (and established that nothing was using it outside of library)

### DIFF
--- a/library/lib/list.php
+++ b/library/lib/list.php
@@ -172,69 +172,15 @@ function listCopy($table, $ids, $field)
 {
     global $translate;
 
-    $hasSubtable = db_tableexists($table.'_content');
     $fieldexists = db_fieldexists($table, $field);
-    $hasChildren = db_fieldexists($table, 'parent_id');
 
     if (!$fieldexists) {
         return [false, $translate['cms_list_copyfailure'], true];
     }
 
-    if ($hasSubtable) {
-        listCopy_multilanguage($ids, $table, $field);
-    } else {
-        listCopy_single($ids, $table, $field);
-    }
+    listCopy_single($ids, $table, $field);
 
     return [true, $translate['cms_list_copysuccess'], true];
-}
-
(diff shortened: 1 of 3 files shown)


library/lib/list.php
```diff
@@ -172,71 +172,17 @@ function listCopy($table, $ids, $field)
 {
     global $translate;
 
-    $hasSubtable = db_tableexists($table.'_content');
     $fieldexists = db_fieldexists($table, $field);
-    $hasChildren = db_fieldexists($table, 'parent_id');
```
HaGuesto: `$hasChildren` has no functionality in this function. Wicked!

library/lib/list.php
```diff
@@ -457,7 +403,7 @@ function addpagemenu($code, $label, $options = [])
 
 function getlistdata($query, $parent = 0)
 {
-    global $table, $settings, $listconfig, $action;
```
HaGuesto: uhhh. This might be dangerous.

HaGuesto: It seems like it should be fine. However, sometimes you have some helper functions which want to access these variables.

PR #185, opened 2019-09-21
Improve performance of beneficiaries page

jamescrowley:
By removing N+1 db query.

Instead of separately querying each row for its parent, we sort the results from the database
and order them such that for each parent, its direct children appear immediately next in the list.

Ideally we'd use a CTE for this, but Google Cloud SQL doesn't support MySQL 8 yet (which is when this became supported) so instead we'll do a poor man's version here and compute on the fly.

In order to preserve arbitrary sorting, we record the index of the row returned from the database, and build a 'path' representing it.

### DIFF
--- a/include/people.php
+++ b/include/people.php
@@ -32,7 +32,7 @@ $table = $action;
 
         $search = substr(db_escape(trim($listconfig['searchvalue'])), 1, strlen(db_escape(trim($listconfig['searchvalue']))) - 2);
 
-        $data = getlistdata('
+        $data = gettreedata('
 		SELECT 
 			IF(DATEDIFF(NOW(),
 			IF(people.parent_id,NULL,GREATEST(COALESCE((SELECT transaction_date 
--- a/include/people.php
+++ b/include/people.php
@@ -32,7 +32,7 @@ $table = $action;
 
         $search = substr(db_escape(trim($listconfig['searchvalue'])), 1, strlen(db_escape(trim($listconfig['searchvalue']))) - 2);
 
-        $data = gettreedata('
+        $data = getlistdata('
 		SELECT 
 			IF(DATEDIFF(NOW(),
 			IF(people.parent_id,NULL,GREATEST(COALESCE((SELECT transaction_date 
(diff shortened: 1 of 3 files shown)


library/lib/list.php
```diff
@@ -455,11 +401,25 @@ function addpagemenu($code, $label, $options = [])
     }
 }
 
-function getlistdata($query, $parent = 0)
+// this is a huge SQL injection risk, as we're forcing consumers
+// to pass $query with all parameters already supplied. WHY?
```
HaGuesto: saw it before, not sure why we are not escaping here. we do it in all other db functions.

library/lib/list.php
```diff
@@ -455,11 +401,25 @@ function addpagemenu($code, $label, $options = [])
...
+function getlistdata($query)
+{
+    return db_array(buildlistdataquery($query));
```
HaGuesto: You can pass a second variable to `db_array``- an `$array`. This is used for the preparation of sql queries.
You should pass that second variable through `getlistdata`, too.

This way at least we do not have to touch this function in the future again and can slowly start to change the SQL queries.

library/lib/list.php
```diff
@@ -522,9 +461,78 @@ function getlistdata($query, $parent = 0)
...
+// the data is sorted such that after any parent all its children immediately follow
+// this is a stable sort, so any other sorting criteria will be preserved
+
+// in addition to the normal row data, the following additional data items are added:
+// for each row, record :
+//   - a 'level' / depth, indexed from 0.
+//       Typically used by the UI components.
+//   - a 'path' of IDs to the ultimate parent, delimited by a /
+//       This is just for visualization/debugging purposes
+//   - a 'path' of to the ultimate parent, but using the original sort order index rather than the Id
```
HaGuesto: cool. never knew how to do this.

library/lib/list.php
jamescrowley: Unfortunately even in the places where it's supported the values are still inserted using string concatenation. There didn't seem much point adding an unused parameter and simple to add later if we decide to address (I'm not envisaging other code using this function anyway given its problems!). Perhaps I can follow up with a separate PR fixing some if that's important?

jamescrowley: Unfortunately even in the places where it's supported the values are still inserted using string concatenation. There didn't seem much point adding an unused parameter and simple to add later if we decide to address (I'm not envisaging other code using this function anyway given its problems!). Perhaps I can follow up with a separate PR fixing some if that's important?

jamescrowley: the way the query function is written, it sticks extra random parameters into the query itself too. Not so straightforward to make 'good' at this point - so pushing out of scope unless you have major objections.

jamescrowley:
@HaGuesto I've updated so that now we're actually avoiding the 'tree' code entirely for the people screen, given there's only one level of hierarchy. So it's just the cms_functions code that uses the 'tree' stuff now. I'd still suggest we look at limiting the amount of data we return, but at least this approach still leaves that path open to us.

library/lib/list.php
HaGuesto: fine for me. There must have been a reason before why they did not do escaping in getlistdata. maybe a question for @barto74 ?

include/people.php
```diff
@@ -70,7 +72,7 @@
 			 	 p.parent_id = people.id AND NOT p.deleted AND p.camp_id = '.$_SESSION['camp']['id'].'
 			 ))
 			' : ' ')
-        .'GROUP BY people.id ORDER BY people.seq');
+        .'GROUP BY people.id ORDER BY IF(people.parent_id,parent.seq + (people.seq / 100000), people.seq)');
```
HaGuesto: @jamescrowley a problem might only come up if our seq number will be bigger than 100000. 
At the moment we are at max seq of 1531 in production. It is 2547 at the Drop in the Ocean instance. However, the if they are non parents the highest one I have seen is 12. Should be fine.

PR #186, opened 2019-09-21
Fix: valid mobile organisation selector request URI

DurkoMatko:
**fix:** mobile organisation selector request URI

(I guess only boxwise gods get to select organisation right? So noone probably noticed this as we don't use mobile :D)

### DIFF
--- a/templates/mobile_selectorganisation.tpl
+++ b/templates/mobile_selectorganisation.tpl
@@ -1,6 +1,6 @@
 <h2 class="page-header">Select an organisation</h2>
 <div class="btn-list">
 	{foreach $organisations as $key=>$organisation}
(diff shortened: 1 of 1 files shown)


DurkoMatko:
@jamescrowley I've did my best to try to create such scenario but couldn't get it. It seems it's always the first query parameter. I guess it wasn't in the past?

jamescrowley:
I think scanning a box could trigger this, no? 

https://staging.boxwise.co/mobile.php?barcode=abc

jamescrowley:
I think it would be safer to just make the template handle both, in case it changes again in future.

DurkoMatko:
@jamescrowley Done.

jamescrowley:
Thanks @durkomatko, much appreciated :)

PR #187, opened 2019-09-22
Feature cypress qr code generation tests

DurkoMatko:
**add:** browser tests for qr labels generation
**add:** seed data for usergroup with no permissions (to test with user who is a member of such group and has no rights)


DurkoMatko:
@jamescrowley You probably won't agree on `cy.wait()` use :D I didn't figure out how to wait till those QRs are generated. Do you maybe have an idea?

DurkoMatko:
Edit: Probably just found a way how to disable chrome web security for only particular tests rather than global (@naphets123 this could also be interesting for you if it ends up working). Will test today in the evening and update if it works. Removed James from reviewers for now.

cypress/integration/3_1_QrCodeGenerationTests.js
```diff
@@ -0,0 +1,109 @@
...
+        labelsCountInputIsVisible();
+        cy.verifyActiveSideMenuNavigation('menu_qr');
+    });
+
+    it('(Desktop) Generate 10 QR codes - small', () => {
+        let numberOfQrs = 10;
+        typeNumberOfLabels(numberOfQrs);
+        uncheckBigLabelsCheckbox();
+        clickMakeLabelsButton();
+        cy.wait(4000);  //no idea how to wait for QRs being generated
```
jamescrowley: Cypress will automatically wait for a condition to be true? - https://docs.cypress.io/guides/core-concepts/retry-ability.html#Commands-vs-assertions - so then it will continue the moment the condition is true, rather than always waiting 4 seconds even if it appears sooner?

jamescrowley: (in theory :D)

jamescrowley:
@DurkoMatko actually, now I'm just confused. We don't have a mobile specific path for generating QR codes, so how is this different from https://github.com/boxwise/dropapp/pull/175 ?

cypress/integration/3_1_QrCodeGenerationTests.js
```diff
@@ -0,0 +1,109 @@
...
+        cy.get("body").then($body => {
+            expect($body.find("div[data-testid='boxlabel-small']").length).to.equal(numberOfQrs);
```
jamescrowley: So weird that we generate a PDF for the large labels but not the small ones.... :/

DurkoMatko:
@jamescrowley re similarity with https://github.com/boxwise/dropapp/pull/175...I had this brief thought yesterday as well..cuz Stephan is doing some QR codes generation as well or? And as you said there is no mobile version..but dunno, the tests seem to be have nice reasonable structure so I've thought I'm just missing something and just continued...did we both with stephan test the same thing? :D Playing with box creation form now, there's no QR code generation in that form

DurkoMatko:
@naphets123 sorry for tagging you in the morning. The potential solution to disable webSecurity just temporary doesn't work. You can try it if you want. It's this line - `Cypress.config('chromeWebSecurity',true);`.....it doesn't work for me though :/

cypress/integration/3_1_QrCodeGenerationTests.js
DurkoMatko: Sorry that was stupid...I've thought that I can get length of elements only if I get body first..dunno why...and it gets the body in the moment where QRs didn't exist at all...now getting already the QR divs and the _then_ block already has them in..This should be fixed now.

cypress/integration/3_1_QrCodeGenerationTests.js
```diff
@@ -67,10 +67,9 @@ describe('QR labels tests - user with rights', () => {
         typeNumberOfLabels(numberOfQrs);
         uncheckBigLabelsCheckbox();
         clickMakeLabelsButton();
-        cy.wait(4000);  //no idea how to wait for QRs being generated
-        cy.get("body").then($body => {
-            expect($body.find("div[data-testid='boxlabel-small']").length).to.equal(numberOfQrs);
-        });
+        cy.get("div[data-testid='boxlabel-small']").then($smallLabels => {
```
jamescrowley: Nice! :)

jamescrowley:
> **Update**: Ok, I've looked into it now...it seems form Stephan is testing checks the box he created and generates QR code with all the fields already set ...it's basically the QR code generation for boxes which already exist in the system...My "Generate QR labels" for creates QR codes with empty values which you can print and use on the fly as the boxes are created. And you have to write values on the QR with pen upon putting the label on the box. Or?

Can we make this more obvious? Should we combine the tests and just add those as a specific scenario?

PR #189, opened 2019-09-23
Fk languages

naphets123:
Added foreign keys, also had to change seeds, since definition/order of cms_users in demodb was different from order in minimaldb


jamescrowley:
@naphets123 I don't quite understand what's happening with the seed data here. Do you remember why it needed changing? (there are conflicts so gonna have to resolve vs the current master)

PR #197, opened 2019-09-25
Feature cypress qr code generation tests

jamescrowley:
Re-opening  #187 due to revert following unexpected CircleCI failure - https://circleci.com/gh/boxwise/dropapp/1438

"Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged."

appears to be some UI interactivity issue with actions that result in PDF generation.


DurkoMatko:
@jamescrowley I think this can be closed as we found the issue was in the data

jamescrowley:
@durkomatko no, the data was a separate issue. After merging this it failed with the error above. After reverting this, it failed with the data issue. So the UI interaction issue remains. May need to tweak the CircleCi config so can easily test without pushing to master?

PR #198, opened 2019-09-27
Add access for non Boxwise Gods to stock_confirm

HaGuesto:


jamescrowley:
Any context on this @HaGuesto ? :)

PR #208, opened 2019-10-03
Rewrite some minor history for schema migrations

jamescrowley:
These have already been applied to production, but make the DiH migration a little easier.

- organisation_id on cms_users is created as nullable. This column gets changed to nullable in the next step anyway, so this stops them all going to 0 instead of null when the column is created
- userlevel FK points to the correct FK table. This FK was incorrect and gets fixed in 20190701120113_fix_usergroups_foreign_key - this stops unnecessary errors being triggered on data.


jamescrowley:
@HaGuesto more a heads up than anything. This combined with https://github.com/jamescrowley/DiHMigration now migrates the last DiH data snapshot successfully

PR #224, opened 2019-10-13
Fix editing of parents

jamescrowley:
Can't use 0 as that doesnt work as a FK. blank and default drop down value now the same, so just cheating and using that as the 'no parent' selector.


jamescrowley:
Did this the wrong way around 🤕

PR #226, opened 2019-10-17
Fix beneficiary list ordering - change to last/first name

jamescrowley:
We're ignoring the seq column as it's just a bit weird, so now prevent allowing to move items around in the list.

Also fixing bugs in this ordering as reported by DiH, due to dodgy logic on my part. We need to order the columns for the parents first, then ensure children are grouped with them, by ordering on their *parents* data first. Ideally I'd write tests for these kinds of things, but we haven't set up integration style tests yet.

### DIFF
--- a/include/people.php
+++ b/include/people.php
@@ -72,7 +72,17 @@ $table = $action;
 			 	 p.parent_id = people.id AND NOT p.deleted AND p.camp_id = '.$_SESSION['camp']['id'].'
 			 ))
 			' : ' ')
-        .'GROUP BY people.id ORDER BY IF(people.parent_id,parent.seq + (people.seq / 100000), people.seq), IF(people.parent_id,1,0)');
+        .'GROUP BY people.id 
+        ORDER BY 
+            -- sort by *parent* seq (or own seq if no parent)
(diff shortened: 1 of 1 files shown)


HaGuesto:
@jamescrowley if you say you prevent the drag and drop, does this mean that you cannot drag a person to become the child of another person anymore?

jamescrowley:
@HaGuesto Hmm, yes - that is right. I'm not sure with the grid we can allow that, with preventing re-ordering?

jamescrowley:
We'd probably have to implement some client side logic to only allow that, which... is gonna be way too much effort. We could leave it enabled, and then they can still move it - and when you refresh the page, the ordering will reset back again.

aerinsol:
This was my point also - mentioned in Slack and Trello

include/people.php
```diff
@@ -14,7 +14,7 @@
         $cmsmain->assign('title', 'Beneficiaries');
 
         listsetting('allowcopy', false);
-        listsetting('allowmove', true);
+        listsetting('allowmove', false);
```
HaGuesto: I tested what you proposed and it works. So, I would just remove this setting and we are good to go from my side.

PR #227, opened 2019-10-18
Removed foreign key for x_people languages to fix the error in listDe…

naphets123:
…lete as the query is searching for b.id but x_people languages does not have an id column


HaGuesto:
close due to https://trello.com/c/riP8UakK

PR #231, opened 2019-10-19
Make FK check in listdelete function conditional

HaGuesto:


HaGuesto:
[Products cannot be deleted because of Foreign key constraint](https://trello.com/c/Wdf6mPCf/327-products-cannot-be-deleted-because-of-foreign-key-constraint)

PR #232, opened 2019-10-21
Fk location id in stock table

HaGuesto:


HaGuesto:
based on wrong branch

PR #236, opened 2019-10-23
Change encoding and collation to utf8 for dih

HaGuesto:


HaGuesto:
[Collation and Encoding from latin1 to utf8 for Drops database](https://trello.com/c/ENhvuSsI/337-collation-and-encoding-from-latin1-to-utf8-for-drops-database)

PR #239, opened 2019-10-25
Correct installation instructions

jamescrowley:


jamescrowley:
Realised we still need to install PHP, as we use composer to install the dependencies. In theory this could all happen within the docker container, but for now I'll reinstate that bit

PR #240, opened 2019-10-26
New tabs in cms users

HaGuesto:


HaGuesto:
[Deleting a user fails to place them in the deactivated status](https://trello.com/c/ct0zbRQj/302-deleting-a-user-fails-to-place-them-in-the-deactivated-status)

PR #242, opened 2019-10-29
Remove all instances of $_SESSION['camp'] in ajax checkout call.

HaGuesto:
Instead of glueing together how the session data is set in an ajax call, I decided to rather remove all references of `$_SESSION['camp']` in the check_out ajax section.
This way we reduce the usage of the session data and hopefully figure out other places where we can replace them.
It was possible since the ajax-call passes data in a POST request from which the camp can be derived.
For security I added a general check if the user has access to the requested camp.

### trello ticket
https://trello.com/c/kgQs1Vah

### sentry issue
DROPAPP-AE

### DIFF
--- a/include/check_out.php
+++ b/include/check_out.php
@@ -51,36 +51,32 @@
         $cmsmain->assign('formelements', $formdata);
         $cmsmain->assign('formbuttons', $formbuttons);
     } else {
-        if ($_POST['do']) {
-            switch ($_POST['do']) {
-                case 'delete':
-                    $ids = explode(',', $_POST['ids']);
-                    list($success, $message, $redirect) = listDelete($table, $ids);
-
-                    break;
-                case 'hide':
-                    $ids = explode(',', $_POST['ids']);
-                    list($success, $message, $redirect) = listShowHide($table, $ids, 0);
-
-                    break;
-                case 'show':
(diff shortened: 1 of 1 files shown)


HaGuesto:
[SQL Error in Production (IHA & R4R)](https://trello.com/c/kgQs1Vah/321-sql-error-in-production-iha-r4r)

include/check_out.php
```diff
@@ -51,36 +51,32 @@
...
             echo json_encode($return);
             die();
         }
 
+        // verify acces if data of a person is requested
+        verify_campaccess_people($_POST['people_id']);
+        verify_deletedrecord('people', $_POST['people_id']);
+
+        // set camp depending on people_id
+        $camp = db_row('
```
DurkoMatko: I know this shouldn't happen, but I'm just wondering what happens if the query fails and returns null or something...should we maybe check whether the $camp resultset has a value? And if it doesn't then silently skip the next steps or maybe return some error message?

HaGuesto: Camp should not be null in any case. If so it will throw the same error as before and in line 68 there is already a check if the user has access.

HaGuesto:
Thanks for the review, Martin!

PR #246, opened 2019-10-29
do not show deleted boxes in Boxes menu

HaGuesto:
I found this bug because the number of boxes on the start page did not agree with the one in the Boxes menu.

### DIFF
--- a/include/stock.php
+++ b/include/stock.php
@@ -31,7 +31,7 @@
 			LEFT OUTER JOIN locations AS l ON l.id = stock.location_id
 			LEFT OUTER JOIN genders AS g ON g.id = p.gender_id
 			LEFT OUTER JOIN sizes AS s ON s.id = stock.size_id
-		WHERE l.camp_id = '.$_SESSION['camp']['id'].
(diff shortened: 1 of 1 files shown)


include/stock.php
```diff
@@ -31,7 +31,7 @@
 			LEFT OUTER JOIN locations AS l ON l.id = stock.location_id
 			LEFT OUTER JOIN genders AS g ON g.id = p.gender_id
 			LEFT OUTER JOIN sizes AS s ON s.id = stock.size_id
-		WHERE l.camp_id = '.$_SESSION['camp']['id'].
+		WHERE (NOT stock.deleted OR stock.deleted IS NULL) AND OR l.camp_id = '.$_SESSION['camp']['id'].
```
DurkoMatko: I think you wanted to use only `AND` here, right?

HaGuesto: yes :)

HaGuesto:
[fix stock query](https://trello.com/c/tu3m9f50/368-fix-stock-query)

PR #247, opened 2019-10-31
Delete boxes cypress helper function

DurkoMatko:
**change:** deleteAllBoxes is a global helper cypress function now and it also makes sure to not delete seed box
**change:** mobile box creation cleans up created boxes


cypress/integration/3_2_Box_creation_desktop.js
```diff
@@ -15,7 +15,7 @@ context("Box_creation_tests", () => {
     })
 
     after(function(){
-        DeleteAllBoxes();
+        cy.deleteAllBoxes();
```
jamescrowley: What's the reason for deleting at the start & the finish @DurkoMatko? What if we just do it at the start?

cypress/support/box-helper-functions.js
```diff
@@ -0,0 +1,24 @@
+const SEED_BOX_COMMENT = "Cypress seed test box";
+
+Cypress.Commands.add("deleteAllBoxes", () => {
```
jamescrowley: 👍

cypress/support/box-helper-functions.js
```diff
@@ -0,0 +1,24 @@
...
+    cy.loginAsVolunteer();
+    cy.visit('/?action=stock');
+    cy.get("input[data-testid='select_all']").click();
+
+    cy.get("body").then($body => {
+        // make sure the seed box will not get deleted!
+        if ($body.text().includes(SEED_BOX_COMMENT)) {
```
jamescrowley: Hah. Maybe we should call it 'deleteAllBoxesExceptSeed' ? :grin:

cypress/integration/3_3_MobileBoxCreation.js
```diff
@@ -106,11 +111,14 @@ describe('Mobile box creation using QR scanning (logged-in user)', () => {
...
             selectLocation(LOCATION3);
             defineItemsCount(itemsCount);
+            writeComment(SAME_ORG_BOX_COMMENT);
             clickNewBoxButton();
             // assertions
             cy.mobileNotificationWithTextIsVisible('contains ' + itemsCount + ' ' + PRODUCT1);
             cy.mobileNotificationWithTextIsVisible('located in ' + LOCATION3);
             checkBoxContent(PRODUCT1, size, LOCATION3, itemsCount);
+            // cleanup
+            cy.deleteAllBoxes();
```
jamescrowley: I guess do this at the start too? Or are there problems with that @DurkoMatko ?

cypress/integration/3_2_Box_creation_desktop.js
DurkoMatko: I just thought it'd be safer to do it twice :D Since we're cleaning up with clicking the DOM, I just somehow don't believe it 100%...if the cleanup fails, tests can potentially start passing even if they shouldn't (cuz there will be boxes from previous runs which will make the assertions pass)..But yeah, running cleanup twice takes more time....should I delete it? Up to you :)

cypress/integration/3_3_MobileBoxCreation.js
DurkoMatko: Hmm true, here I've forgotten on my "double cleanup" policy :D Dunno, so should we rather delete twice to be more "robust"? :D Or once is enough? In the comment above, you seemed you  think once is enough. So not sure if I should 2nd one here now.

cypress/support/box-helper-functions.js
```diff
@@ -0,0 +1,24 @@
...
+            cy.get('tr').contains("Martin test comment").parent().parent().within(()=>{
+                cy.get("input[type='checkbox']").uncheck();
+            });
+        }
+        return cy.get('button:visible');
+    })
+    .then($button => {
+        if ($button.text().includes("Delete")) {
+            cy.log("Delete button is visible")
```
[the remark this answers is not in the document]
DurkoMatko: asnwered in the first PR...cy.log is just a log message, it doesn't have any affect on the flow of the tests...I've copied if from your conditional cleanup of boxes so that we're consistent with cleanup logging :D

cypress/support/box-helper-functions.js
DurkoMatko: Done.

cypress/integration/3_2_Box_creation_desktop.js
DurkoMatko: Ok, cleanup is done only in afterEach hook.

PR #253, opened 2019-11-02
Disallow escaping in smarty

DurkoMatko:
change: escape html smarty configuration turned on
add: nofilter smarty flag added at several places to not escape template variables


DurkoMatko:
[Disallow escaping in Smarty #medium](https://trello.com/c/65NyN2za/290-disallow-escaping-in-smarty-medium)

PR #254, opened 2019-11-03
unique constraints for boxid and code/legacy in stock and qr tables

cohenlea:


cohenlea:
[Implement unique constraint for box_id in stock table #tiny](https://trello.com/c/MHlRZR3I/348-implement-unique-constraint-for-boxid-in-stock-table-tiny)

cohenlea:
[Implement unique constraint for hash-legacy-combination in qr-db-table #small](https://trello.com/c/KEtU8lbK/349-implement-unique-constraint-for-hash-legacy-combination-in-qr-db-table-small)

cohenlea:
> @cohenlea I ran the migrations and they worked 🎉
> 
> However, could you rewrite them in the syntax of [phinx indexes](http://docs.phinx.org/en/latest/migrations.html#working-with-indexes).
> 
> This way we could undo this migration when running backwards.

Yeah will do! In postgres indexes and constraints are different things, didn't realize I could use an index in mysql.

PR #255, opened 2019-11-03
Fix white page after beneficiary is created

DurkoMatko:
**change**: redirect to beneficiary page after adding new beneficiary from menu list
**add**: assertion in add beneficiary cypress test


DurkoMatko:
[White page after beneficiary is created #small](https://trello.com/c/Is7o2HQq/355-white-page-after-beneficiary-is-created-small)

PR #259, opened 2019-11-05
Ajax to delete test data

HaGuesto:
This is a possible solution for a function to delete test data after it has been created.
You need to pass the database ids and the name of the database table to remove the data from the database.

However, is it possible to get the database ids of the records you created during tests?

Is that useful? Would you do it differently?


HaGuesto:
[Ajax-functions to delete db entries which were created in tests](https://trello.com/c/PsIxMguB/373-ajax-functions-to-delete-db-entries-which-were-created-in-tests)

library/ajax/testdbdelete.php
```diff
@@ -0,0 +1,38 @@
...
+        LEFT JOIN cms_usergroups g ON u.cms_usergroups_id = g.id
+        WHERE g.organisation_id = 100000000'
+    );
+
+    // Test if user is test user and the ids can be deleted
+    // on production environment the users never have an id bigger than 100000000
+    $return = ['success' => true];
+    foreach ($ids as $id) {
+        if ($_SESSION['user']['id'] >= 100000000 && in_array($id, $allowed[$_POST['table']])) {
+            listRealDelete($_POST['table'], $ids);
```
jamescrowley: this is a SQL injection risk, because listRealDelete doesn't escape the table name :)

library/ajax/testdbdelete.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+    $ids = explode(',', $_POST['ids']);
```
jamescrowley: as a belt and braces, perhaps could immediately stopping script execution unless the host name is staging.boxwise.co or localhost? could also check $settings['db_database'] and only run if its dropapp_staging or dropapp_dev ?

naphets123:
To james comment, i would propose to use the email, since it has to be unique in our database.
We can then query for the id and use listRealDelete

HaGuesto:
@naphets123 and @DurkoMatko my question was if you think this function is helpful at all.

I mean you will have to pass the database ids of the created boxes, users or beneficiaries to delete them. Do we actually manage that in Cypress. Can you collect the db ids of the boxes, users or beneficiaries you created?

library/ajax/testdbdelete.php
HaGuesto: With the new commit, the table input is validated. So, should be resolved

library/ajax/testdbdelete.php
HaGuesto: done

jamescrowley:
@HaGuesto just a thought - would it be simpler / make sense to just trigger a re-seed of the data?

HaGuesto:
@naphets123 I added functionality so you can delete a user by specifying emails.

PR #260, opened 2019-11-10
Prevent escaping of mobile notifications and box history

DurkoMatko:
**change:** Prevent escaping of box history
**delete:** html elements passed via GET in mobile notifications
**add:** support for anchor links in mobile notifications without passing the whole DOM in the URL query (creatable by 3 parameters - messageAnchorText, messageAnchorTarget and messageAnchorTargetValue)

Example:
`&messageAnchorText=Go%20back%20to%20this%20box&messageAnchorTarget=boxid&messageAnchorTargetValue=100000003`

image


DurkoMatko:
[html escaping of Box history on mobile screen #tiny](https://trello.com/c/uxhH3K83/377-html-escaping-of-box-history-on-mobile-screen-tiny)

jamescrowley:
@durkomatko two things:
- your PR says escape but if I'm understanding correctly this is explicitly *not* escaping in these cases?
- am I right in understanding these parameters are set from the query string? If so by not escaping, you can directly inject JavaScript etc and thus an XSS vulnerability. This was the kind of scenario which we need to force escaping for. Can we remove the html passed in this parameter, or find some other way of triggering the message?

Let me know if I'm mistaken though - this is just a mobile phone review!

DurkoMatko:
@jamescrowley 

- Hah, yupp you're right, it's preventing smarty escaping. My bad
- Right again :D Removing html would work if we are alright with losing font styling (boldness)...but there's also an anchor element constructed which takes user back to the box overview.

image

I can think of some other approach (maybe add the anchor tag as optional to the the template) ..but in general, we trigger all the notifications (messages) by redirecting with the message defined in the URL. Maybe we could add some toast notification lib and use that?...but not sure if it's worth the time if this is the only case where current approach is causing the issues.

templates/mobile.tpl
```diff
@@ -45,7 +45,16 @@
 				</div>
 			{/if}		
 		</header>
-		{if $data['message']}<div class="message {if $data['warning']}warning{/if}">{$data['message']}</div>{/if}
+		{if $data['message']}
+			<div class="message {if $data['warning']}warning{/if}">
+				{$data['message'] nofilter}
+				{if $data['messageAnchorText'] && $data['messageAnchorTarget'] && $data['messageAnchorTargetValue']}
+					<a href="?{$data['messageAnchorTarget']}={$data['messageAnchorTargetValue']}">
+						{$data['messageAnchorText'] nofilter}
```
jamescrowley: I think you don't need the nofilter here and above now? That's the 'dangerous' element with respect to XSS

jamescrowley:
Yeah the whole pass the message in the query string pattern is broken for several reasons - when you're taken to the page it's shown - and then if you're on a page with one of those weird client side reload JavaScript triggers then the message displays again even though you've performed a completely different action. :/

templates/mobile.tpl
DurkoMatko: Don't you need it also in case when you're specifying html inline as a part of the string? I've written u on slack.

DurkoMatko: Ok. Removed the nofilter from messageAnchorText variable...above for message it's still needed. If it's not there, the inline html string gets rendered..

HaGuesto:
@jamescrowley side info. The client-side redirect is only possible on the desktop environment.
On mobile the redirect is executed from the server-side and the warning and message is passed in the url.

templates/mobile.tpl
```diff
@@ -45,7 +45,16 @@
 				</div>
 			{/if}		
 		</header>
-		{if $data['message']}<div class="message {if $data['warning']}warning{/if}">{$data['message']}</div>{/if}
+		{if $data['message']}
+			<div class="message {if $data['warning']}warning{/if}">
+				{$data['message'] nofilter}
```
jamescrowley: This one shouldn't have nofilter either @DurkoMatko, I think - it comes from the $_GET too ?

templates/mobile.tpl
jamescrowley: Sorry, just seen this. Where does the HTML come from? If we have nofilter here, then we haven't really solved anything regarding XSS,b ecause the $data[message] comes from the GET parameter.

DurkoMatko: The HTML comes from `move.php` file. It's inline defined in a string where we redirect to. **There's no new user input in it.** It contains names of the locations and products which are custom. So those could contain some injection scripts - but they're not new. I guess they've already been checked/escaped/whatever upon creation right?

HaGuesto:
@DurkoMatko @jamescrowley 

@DurkoMatko I am very sorry, but I found another four places where we should prevent escaping.
I created a trello card for each of them.
1. https://trello.com/c/J44LwGJ6 (most similiar to what you have done already)
2. https://trello.com/c/MqkhEghB
3. https://trello.com/c/4hLSgnHx
4. https://trello.com/c/pydMp0zb

templates/mobile.tpl
jamescrowley: @DurkoMatko the problem is that while we put 'safe' HTML in the redirect path, there's nothing to stop someone crafting a malicious URL instead. So for instance,

https://app.boxwise.co/mobile.php?message=%3Cscript%3Ealert(%27hello%27)%3C/script%3E

And once you can do that, you can do this: https://pentest-tools.com/blog/xss-attacks-practical-scenarios/

DurkoMatko: Oh I understand now..so should we just drop the HTML parts in our redirect and therefore there won't be a need to use `nofilter` for the message?

jamescrowley: @DurkoMatko :) I think that's probably the best approach.

jamescrowley:
@DurkoMatko regarding those ^^ unfortunately the same will apply - any smarty data that comes from the db or a get/request parameter would also have to be encoded to avoid XSS, so rather than using 'nofilter', instead ideally we'd have to adjust the code so it doesn't pass raw HTML through. But may be easier said than done. // @HaGuesto

DurkoMatko:
@HaGuesto the first one is already a part of this PR.

templates/mobile.tpl
DurkoMatko: done

templates/mobile.tpl
DurkoMatko: done

DurkoMatko: will look at the issues @HaGuesto linked here in another branch. Just ran a quick regex on php files and there seems to be couple of more places where this needs to be checked...

HaGuesto: perfect.

PR #262, opened 2019-11-12
Clean up dev seed

HaGuesto:
- Added [Faker library](https://github.com/fzaninotto/Faker)
- The dev seed has now two organisations - one with one base and one with two bases.
- Each DemoDB base has about 1000 people, 100 users, 3000 boxes, 200 products seeded.
- For Cypress I only added seeds for people (10000) and stock (20000) to test the performance.
- The app is getting slower with this amount of data in it. :)


HaGuesto:
[Clean up demo db](https://trello.com/c/0QAV3pH3/356-clean-up-demo-db)

HaGuesto:
## testing instructions
We might have to make the seed smaller for performance in the dev environment.
1.) please checkout this PR 
```
git fetch origin
git checkout -b clean_up_dev_seed origin/clean_up_dev_seed
git merge master
```
2. Install the faker library
```
composer install
```
3. Run the seed script
```
vendor/bin/phinx seed:run -e development
```

HaGuesto:
I added now 3500 transactions.

DurkoMatko:
@HaGuesto,  what exactly is a "dev base"? dev "camp"? What does the dev part stand for? Cuz we don't have a dev seed, do we? There's minimaldb, demodb and testdb, right? I'm just trying to figure out whether my db got seeded correctly. Cuz I only have 115 users in my DB. And 3 organisations..shouldn't I have more users than just 115?

And the app is actually quite slower with this ammount of data :/

HaGuesto:
> @HaGuesto, what exactly is a "dev base"? dev "camp"? What does the dev part stand for? Cuz we don't have a dev seed, do we? There's minimaldb, demodb and testdb, right? I'm just trying to figure out whether my db got seeded correctly. Cuz I only have 115 users in my DB. And 3 organisations..shouldn't I have more users than just 115?
> 
> And the app is actually quite slower with this ammount of data :/

@DurkoMatko I do not know what you mean with "dev base". I cannot find that expression in the code.

115 users make sense. I can increase it if you want.

About the performance we should actually check with @jamescrowley. I'm less worried about MYSQL queries than our frontend.

DurkoMatko:
@HaGuesto Hah I don't know what it is either :D You're mentioning it in the PR description, that's why I'm asking :D in the third bullet point

HaGuesto:
> @HaGuesto Hah I don't know what it is either :D You're mentioning it in the PR description, that's why I'm asking :D in the third bullet point

Ah I see. Yes the dev base was stupid and just means all bases in DemoDb seed

HaGuesto:
> I have an issue after seeding with this new seed. As weird as it sounds, when I navigate to "Add Beneficiary", it looks like the edit form with the user with refugee ID 8993 :D When I drop the DB, switch to master, reseed and then just refresh the browser - the issues is gone and I see normal "Add beneficiary" title. If I go back to this branch and reseed & refresh the browser again, issue shows again (same refugee ID, different name - probs cuz of random faker data)
> 
> Selection_434
> 
> Selection_432
> 
> I've deleted cache in my browser etc..not sure what's wrong...can someone confirm this bug or is it just my PC? :D
> 
> Can this be related to the bug of couple of weeks ago where when we navigated to "add beneficiary", manage beneficiaries was shown as selected in the left navigation panel?

happens to me too. super weird.

HaGuesto:
Found the bug. The seed generates a person with db id=0. This one is loaded when a beneficiary is added.

HaGuesto:
> Found the bug. The seed generates a person with db id=0. This one is loaded when a beneficiary is added.

should be fixed now

HaGuesto:
I reduced the Cypress seed @cohenlea. Could you check if the Boxes menu still times out?

HaGuesto:
[clean up dev seed](https://trello.com/c/JW5ExBcu/389-clean-up-dev-seed)

cohenlea:
> I reduced the Cypress seed @cohenlea. Could you check if the Boxes menu still times out?

Works now! :)

PR #264, opened 2019-11-12
changed GET method to SESSION as to prevent failure during export of …

naphets123:
…large amounts of data

### DIFF
--- a/include/people_export.php
+++ b/include/people_export.php
@@ -4,7 +4,7 @@ $result = db_query(
     '
 	SELECT p.*, DATE_FORMAT(FROM_DAYS(DATEDIFF(NOW(), p.date_of_birth)), "%Y")+0 AS age
 	FROM people AS p
-	WHERE camp_id = :campid AND (NOT deleted OR deleted IS NULL) '.($_GET['ids'] ? 'AND p.id in ('.$_GET['ids'].')' : ' AND FALSE'),
+	WHERE camp_id = :campid AND (NOT deleted OR deleted IS NULL) '.($_SESSION['export_ids'] ? 'AND p.id in ('.$_SESSION['export_ids'].')' : ' AND FALSE'),
     ['campid' => $_SESSION['camp']['id']]
 );
 $keys = ['container' => $_SESSION['camp']['familyidentifier'], 'firstname' => 'Firstname', 'lastname' => 'Lastname', 'gender' => 'Gender', 'age' => 'Age'];
--- a/include/people_export.php
+++ b/include/people_export.php
@@ -1,11 +1,18 @@
 <?php
 
+//Create array with the export_ids_people in it
+$export_ids_array = explode(',', $_SESSION['export_ids_people']);
+//Create a list of placeholders ? the same length as export ids given
+$id_pars = str_repeat('?,', count($export_ids_array) - 1).'?';
+//Put camp id as first element in the list
+array_unshift($export_ids_array, $_SESSION['camp']['id']);
+
 $result = db_query(
     '
 	SELECT p.*, DATE_FORMAT(FROM_DAYS(DATEDIFF(NOW(), p.date_of_birth)), "%Y")+0 AS age
 	FROM people AS p
-	WHERE camp_id = :campid AND (NOT deleted OR deleted IS NULL) '.($_SESSION['export_ids_people'] ? 'AND p.id in ('.$_SESSION['export_ids_people'].')' : ' AND FALSE'),
-    ['campid' => $_SESSION['camp']['id']]
+	WHERE camp_id = ? AND (NOT deleted OR deleted IS NULL) '.($_SESSION['export_ids_people'] ? 'AND p.id in ('.$id_pars.')' : ' AND FALSE'),
+    $export_ids_array
 );
 unset($_SESSION['export_ids_people']);
 $keys = ['container' => $_SESSION['camp']['familyidentifier'], 'firstname' => 'Firstname', 'lastname' => 'Lastname', 'gender' => 'Gender', 'age' => 'Age'];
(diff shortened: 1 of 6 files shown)


naphets123:
https://trello.com/c/AEh5qwmR/384-error-on-export-of-large-amounts-of-data-due-to-get-request

include/people_export.php
```diff
@@ -4,7 +4,7 @@
     '
 	SELECT p.*, DATE_FORMAT(FROM_DAYS(DATEDIFF(NOW(), p.date_of_birth)), "%Y")+0 AS age
 	FROM people AS p
-	WHERE camp_id = :campid AND (NOT deleted OR deleted IS NULL) '.($_GET['ids'] ? 'AND p.id in ('.$_GET['ids'].')' : ' AND FALSE'),
+	WHERE camp_id = :campid AND (NOT deleted OR deleted IS NULL) '.($_SESSION['export_ids'] ? 'AND p.id in ('.$_SESSION['export_ids'].')' : ' AND FALSE'),
```
HaGuesto: could you try to passing the exportids like the campid?

include/stock_export.php
```diff
@@ -7,7 +7,7 @@
     LEFT OUTER JOIN locations AS l ON l.id = boxes.location_id
     LEFT OUTER JOIN genders AS g ON g.id = p.gender_id
     LEFT OUTER JOIN sizes AS s ON s.id = boxes.size_id 
-    WHERE l.camp_id = :campid AND (NOT boxes.deleted OR boxes.deleted IS NULL) '.($_GET['ids'] ? 'AND boxes.id in ('.$_GET['ids'].')' : ' AND FALSE'),
+    WHERE l.camp_id = :campid AND (NOT boxes.deleted OR boxes.deleted IS NULL) '.($_SESSION['export_ids'] ? 'AND boxes.id in ('.$_SESSION['export_ids'].')' : ' AND FALSE'),
```
HaGuesto: could you try to passing the exportids like the campid?

include/people_export.php
```diff
@@ -4,7 +4,7 @@
...
     ['campid' => $_SESSION['camp']['id']]
 );
```
HaGuesto: I would add `unset($_SESSION['export_ids'])` here.

include/stock_export.php
```diff
@@ -7,7 +7,7 @@
...
     ['campid' => $_SESSION['camp']['id']]
 );
```
HaGuesto: I would add `unset($_SESSION['export_ids'])` here.

naphets123:
@HaGuesto : i see, whouldn't giving them seperate names prevent this exploit? export_boxes_ids, export_stock_ids .... we could do the unset too

include/people_export.php
naphets123: do you mean by like [campid => $_SESSION['camp_id'], exportids=> $_SESSION['export_ids'] ?

HaGuesto: yes, exactly

HaGuesto:
> @HaGuesto : i see, whouldn't giving them seperate names prevent this exploit? export_boxes_ids, export_stock_ids .... we could do the unset too

Yes, it would. Do it :)

jamescrowley:
@HaGuesto @naphets123 apologies if this has already been discussed, but is there any way we can achieve this without relying on session data again? It just seems like we're setting ourselves up for more strange bugs, given the existing issues we have with SESSION data being used to manage state. Could we perhaps avoid the redirect and do the export immediately while we have the $_POST data? Or ensure the data is POSTed to the export endpoint in the first place?

HaGuesto:
> @HaGuesto @naphets123 apologies if this has already been discussed, but is there any way we can achieve this without relying on session data again? It just seems like we're setting ourselves up for more strange bugs, given the existing issues we have with SESSION data being used to manage state. Could we perhaps avoid the redirect and do the export immediately while we have the $_POST data? Or ensure the data is POSTed to the export endpoint in the first place?

@naphets123 I talked to @jamescrowley and using the SESSION variable is ok and the fastest and easiest solution for now.

include/people_export.php
naphets123: In this case we couldn't use the  Syntax $_SESSION['export_ids'] ? Option A : Option B
So we would add some additional checks at another point

HaGuesto:
@naphets123 missing onw commit to escape the SQL queries.

naphets123:
Done

include/products_export.php
```diff
@@ -2,17 +2,25 @@
...
+
 $result = db_query(
     'SELECT p.*, sg.label AS sizegroup, g.label AS gender, p.value AS drops, COALESCE(SUM(s.items),0) AS items, pc.label AS category 
 		FROM products AS p
 		LEFT OUTER JOIN genders AS g ON g.id = p.gender_id
 		LEFT OUTER JOIN sizegroup AS sg ON sg.id = p.sizegroup_id
 		LEFT OUTER JOIN stock AS s ON s.product_id = p.id AND NOT s.deleted '.($locations ? ' AND s.location_id IN ('.$locations.')' : '').'
 		LEFT OUTER JOIN product_categories AS pc ON p.category_id = pc.id
-		WHERE (NOT p.deleted OR p.deleted IS NULL) AND camp_id = :camp_id '.($_GET['ids'] ? 'AND p.id IN ('.$_GET['ids'].')' : 'AND FALSE').'
+		WHERE (NOT p.deleted OR p.deleted IS NULL) AND camp_id = ? '.($_SESSION['export_ids_products'] ? 'AND p.id IN ('.$id_pars.')' : 'AND FALSE').'
```
HaGuesto: wicked :)

Does this mean that a list of ids cannot be passed as a variable to db_query, e.g.
```
camp_id = :camp_id AND p.id IN :id_list
```

Could you send me a link to the doc where you found that solution?

HaGuesto: https://phpdelusions.net/pdo#in

PR #265, opened 2019-11-12
Load Session Data only when session is checked

HaGuesto:


HaGuesto:
[Redirect to start page or login if session data is not complete.](https://trello.com/c/pDKlwsw5/361-redirect-to-start-page-or-login-if-session-data-is-not-complete)

library/lib/session.php
```diff
@@ -58,27 +62,36 @@ function checksession()
...
         if (isset($_COOKIE['autologin_user'])) { // a autologin cookie exists
             $user = db_row('SELECT * FROM cms_users WHERE email != "" AND email = :email AND pass = :pass', ['email' => $_COOKIE['autologin_user'], 'pass' => $_COOKIE['autologin_pass']]);
-            if ($user) {
-                loadSessionData($user);
-                db_query('UPDATE cms_users SET lastlogin = NOW(), lastaction = NOW() WHERE id = :id', ['id' => $_SESSION['user']['id']]);
-
-                return $result;
+            if ($user) { // the cookie is valied
+                // Check if account is not expired
+                $in_valid_dates = check_valid_from_until_date($user['valid_firstday'], $user['valid_lastday']);
```
jamescrowley: Can we somehow simplify this or the flow above so we don't have to duplicate this? like once we've fetched the $user row in both scenarios?

library/lib/session.php
```diff
@@ -172,17 +183,57 @@ function createPassword($length = 10, $possible = '23456789AaBbCcDdEeFfGgHhijJkK
...
+        // Why ajax?
+        // It can happen that an org does not have internet for example in the checkout. The internet connection is checked before a form is submitted. If no internet connection is detected the entries from the form are kept. To not assign the delayed form submission to a wrong camp it is excepted.
+        $_SESSION['camp'] = reset($camplist);
+    }
+
+    if (isset($_SESSION['organisation'], $_SESSION['camp'])) {
+        return true;
+    }
+
+    return false;
```
jamescrowley: what does it (practically) mean when this is false?

library/lib/session.php
```diff
@@ -172,17 +183,57 @@ function createPassword($length = 10, $possible = '23456789AaBbCcDdEeFfGgHhijJkK
...
+            WHERE ug.id = :id AND (NOT ug.deleted OR ug.deleted IS NULL)', ['id' => $user['cms_usergroups_id']]);
+        $_SESSION['organisation'] = db_row('
+            SELECT * 
+            FROM organisations 
+            WHERE id = :id AND (NOT organisations.deleted OR organisations.deleted IS NULL)', ['id' => $_SESSION['usergroup']['organisation_id']]);
+    }
+    $camplist = camplist();
+    if (1 == count($camplist)) {
+        // if user has only access to one camp then set the information
+        $_SESSION['camp'] = reset($camplist);
```
jamescrowley: Do we actually need to use the reset function, or can we just do $camplist[0] ? Potentially more explicit than relying on the array's internal pointer?

jamescrowley:
👍 to aiming to simplify our lovely Session handling, @HaGuesto :grin:

library/lib/session.php
HaGuesto: tried that. Unfortunately, `camplist()` creates an array with the camp_id as key. I did not see another way to access the first (and only element in `$camplist`

library/lib/session.php
HaGuesto: will do

library/lib/session.php
HaGuesto: check line 31, 87 and 250

library/lib/session.php
HaGuesto: should be done

library/lib/session.php
jamescrowley: I meant in plain english :grin: *why* does this scenario happen?

HaGuesto: I tried to make it clearer in the code and explained the cases in comments.

HaGuesto:
@jamescrowley please review again.

library/lib/session.php
```diff
@@ -172,17 +174,66 @@ function createPassword($length = 10, $possible = '23456789AaBbCcDdEeFfGgHhijJkK
     return $password;
 }
 
+// session data requires user, usergroup, organisation, camp
+//usergroup is optional for Boxwise Gods
 function loadSessionData($user)
```
jamescrowley: @HaGuesto sorry for the hold up revewing this, but I'd really like to work on this function to improve it's readability.

library/lib/session.php
```diff
@@ -172,17 +174,66 @@ function createPassword($length = 10, $possible = '23456789AaBbCcDdEeFfGgHhijJkK
...
+            WHERE id = :id AND (NOT organisations.deleted OR organisations.deleted IS NULL)', ['id' => $_SESSION['usergroup']['organisation_id']]);
+    }
+
+    // ------------ load camp
+    $camplist = camplist();
+    if (1 == count($camplist)) { // an organisation has only one camp or a user has only access to one camp
+        $_SESSION['camp'] = reset($camplist);
+    } elseif (isset($_GET['camp'], $camplist[$_GET['camp']])) { // the camp is specified in url and the user has access to it
+        $_SESSION['camp'] = $camplist[$_GET['camp']];
+    } elseif ($user['is_admin'] && isset($_GET['camp']) && !isset($camplist)) { // the user is a Boxwise God and camplist is not set ($_SESSION['organisation'] is not set) and the camp is specified in the url
```
jamescrowley: why/how does $camplist become not set?

library/lib/session.php
HaGuesto: @jamescrowley No worries I understand.
Could you tell me what is missing?

library/lib/session.php
HaGuesto: `camplist()` function is defined here
https://github.com/boxwise/dropapp/blob/84cdded62ddddc3f9bcb2f8020f7320caa0bdeb6/library/functions.php#L36

HaGuesto: The query returned is 
```
		SELECT c.* 
		FROM camps AS c'.($_SESSION['user']['is_admin'] ? '' : ', cms_usergroups_camps AS x').'
		WHERE (NOT c.deleted OR c.deleted IS NULL) AND c.organisation_id = :organisation_id'.($_SESSION['user']['is_admin'] ? '' : ' AND x.camp_id = c.id AND x.cms_usergroups_id = :usergroup_id').'
		ORDER BY c.seq', $parameters, false, true);
```

HaGuesto: camplist is not set if
1. your base got deleted/deactivated,
2. the variable `$_SESSION['organisation'] is not set,
3. your user group has access to no base.

HaGuesto: Case 2 is the biggest one to worry about, because it happens the most (when the session of a Boxwise God expires).

library/lib/session.php
jamescrowley: I just find it very hard to understand the scenarios we're dealing with. Obviously not helped by the ways these are changed elsewhere. 

The conditions in the set of if statements are quite hard to understand, at least for me - it should be pretty obvious which if case we'd fall into? Just wondering if we could group or simplify the cases further?

HaGuesto: just to clarify. it is probably just about the camp session data right?

HaGuesto: alright, I see two possible solutions to further simplify this:

1. We can change the `camplist()`-function so it never returns an empty array. For Boxwise Gods this would mean that they could always choose from all existing camps and directly move between them. In case the session expires it would always set it to the first camp in the database.

2. We assign every admin to a user group and organisation. This way `camplist` could also not return any empty array. This way, Boxwise Gods would have a default account they would be forwarded to once the Session expires.

HaGuesto: @jamescrowley

jamescrowley: 👍 , not quite sure on pros and cons of (1) vs (2) but suspect (1) will be easier to try first without db changes?

HaGuesto:
I updated this PR. Now the function `camplist()` cannot return an empty array anymore. 
Hence, I was able to simplify `loadsession()` further.

jamescrowley:
@HaGuesto just did a little bit of clicking around, and got 

Exception: $_SESSION[organisation] or $_SESSION[camp] is not set!

this was the sequence:
http://localhost:8100/?action=start
http://localhost:8100/?action=people_add
http://localhost:8100/?action=start&organisation=100000000
http://localhost:8100/?action=people_add&organisation=2

HaGuesto:
I cannot reproduce it and all tests run through on my side. If it is an error, it is only possible for Boxwise God. I would merge it for now since the Exception you have found is still more clear than the exceptions we get until now when the organisation or camp is not set.

PR #267, opened 2019-11-14
Mobile header

cohenlea:


cohenlea:
[Overflow of header if organisation and camp name is too long on mobile.](https://trello.com/c/XTC9ZrZI/363-overflow-of-header-if-organisation-and-camp-name-is-too-long-on-mobile)

cohenlea:
[camp name is not shown in mobile header for users if they only have access to one camp #small](https://trello.com/c/qVDNttIF/364-camp-name-is-not-shown-in-mobile-header-for-users-if-they-only-have-access-to-one-camp-small)

cohenlea:
The logic is a bit confusing, not sure if the users will notice/care that we sometime show only the org, sometimes both, and sometimes only the camp(s)... maybe we should just always show only the camp(s)? At the moment (after my changes) it's 1) Only the Org name if the combined name is > 34 characters or they only have one camp, 2) Both names if they have multiple camps and the combined name is < 34 characters, 3) Only the camp if the combined name is too long and they have multiple camps.

HaGuesto:
@cohenlea First of all amazing! Without any help you wrote php code and changed the template.

Second, lol. Sorry for giving you such a weird task. I did not really realize that we are doing it once this way and then the otherway :see_no_evil: 

So, yes I agree. It makes sense to always show the name of the camp and if the length allows it to show the organisation name.

If I think of it. The name of organisation is only important for Boxwise Gods to know.
W'll need to talk to Roanna to finalize this.

PR #268, opened 2019-11-14
Restricted date selection in user dates starting from today

naphets123:


HaGuesto:
[User Management: Restrict "Valid From" Date to CurrentDay #tiny](https://trello.com/c/Ns2vEbe4/375-user-management-restrict-valid-from-date-to-currentday-tiny)

PR #270, opened 2019-11-14
2_6 user menu tests

cohenlea:


cohenlea:
[Cypress: User Mgmt. 2.6 through 2.6.2 #medium](https://trello.com/c/el5rF45C/342-cypress-user-mgmt-26-through-262-medium)

cohenlea:
Uses new seed data from https://github.com/boxwise/dropapp/pull/262 and some tests will fail until that pr is merged

cypress/integration/2_6_UserMenu.js
```diff
@@ -0,0 +1,69 @@
...
+        // check that admin can see roles Coordinator and User
+        CheckForElementByText("td[class='list-level- list-column-usergroup'] div", "TestUserGroup_Coordinator");
+        CheckForElementByText("td[class='list-level- list-column-usergroup'] div", "TestUserGroup_User");
+
+        // login as coodrinator; check that coordinator can see roles User but not Admin
+        cy.visit('/?logout=1');
+        cy.loginAsCoordinator();
+        cy.visit('/?action=cms_users');
+        CheckForElementByText("td[class='list-level- list-column-usergroup'] div", "TestUserGroup_User");
+        CheckElementDoesNotExist("tr[id='row-100000000']");
```
DurkoMatko: Is this checking that TestUserGroup_Coordinator rows don't exist? Maybe it'd be better to do this?

`cy.get("tr").contains('TestUserGroup_Coordinator').should('not.exist');`

I'm not sure where does that id come from but I wouldn't trust it to always be the same. It looks like DB id of this user group from test seed..In that case it will work like it is...but dunno, I find this another option bit more readable. Up to you :)

cypress/integration/2_6_UserMenu.js
```diff
@@ -0,0 +1,69 @@
...
+    it("2_6 Check for list elements", () => {
+        CheckForElementByTypeAndTestId("a.active", "active_pending");
+        CheckForElementByTypeAndTestId("a", "expired");
+        CheckForElementByTypeAndTestId("a", "deactivated");
+        CheckForElementByTypeAndTestId("input", "select_all");
+        CheckForElement("a[href='?action=cms_users_edit&origin=cms_users']");
+        CheckElementNotEmpty("td[class='list-level- list-column-email'] div[class='td-content']");
+
+        // check that admin can see roles Coordinator and User
+        CheckForElementByText("td[class='list-level- list-column-usergroup'] div", "TestUserGroup_Coordinator");
```
DurkoMatko: btw if you check other test files, we usually define user/group/product names as consts on the top of the files. So if the seed from some reason changes, we can just change the "headers" of the test files and don't have to search for them in all the tests. Can you maybe add those const variables?

cypress/integration/2_6_UserMenu.js
```diff
@@ -0,0 +1,69 @@
...
+
+    it("2_6_1 Check for list elements for single user edit page", () => {
+        ClickOnElement("td[class='controls-front list-level- list-column-naam'] a[class='td-content-field']");
+        CheckForElement("div[id='div_naam']");
+        CheckForElement("div[id='div_email']");
+        CheckForElement("div[id='div_cms_usergroups_id']");
+        CheckForElement("div[id='div_valid_firstday']");
+        CheckForElement("div[id='div_valid_lastday']");
+        CheckForElement("div[id='div_lastlogin']");
+        CheckForElement("div[class='created light small']");
```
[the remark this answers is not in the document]
cohenlea: Hey @DurkoMatko thanks for the feedback. I cleaned up the strings and added data-testids, also added the option for this attribute in cms_form_created.tpl and cms_form_info.tpl - take a look and let me know if this works.

cypress/integration/2_6_UserMenu.js
DurkoMatko: Solved.

cypress/integration/2_6_UserMenu.js
DurkoMatko: Solved.

cypress/integration/2_6_UserMenu.js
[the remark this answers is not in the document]
DurkoMatko: Nice, looks good! I actually never added any data-testid to template myself :D I always just politely ask Hans to add it for me :D

PR #261, opened 2019-11-11
Add testid option in page tabs for Cypress

HaGuesto:
When using the helper function `addpagemenu` the testid can be passed as an option now.

PR #271, opened 2019-11-17
Use data-testid to navigate user tabs in cypress tests

DurkoMatko:
**change:** use data-testid to navigate user tabs in cypress tests (supported since Hans added the support in https://github.com/boxwise/dropapp/pull/261)

`data-testid` attributes are added as a part of Lea's PR here -  https://github.com/boxwise/dropapp/pull/270 ...so her PR needs to be merged to make this work)


HaGuesto:
@DurkoMatko when you approved and merged #270 from @cohenlea, please also merge this PR.

PR #272, opened 2019-11-19
QR-codes added without existing box are recorded as 'New Qr-code crea…

naphets123:
…ted'


naphets123:
[Track the generation of QR-codes in the history table #small](https://trello.com/c/Xv6d4q2O/283-track-the-generation-of-qr-codes-in-the-history-table-small)

pdf/qr.php
```diff
@@ -51,6 +51,8 @@
                 $from['int'] = $box['qr_id'];
             }
             simpleSaveChangeHistory('stock', $labels[$i], 'New Qr-code assigned.', $from, ['int' => $id]);
+        } else {
+            simpleSaveChangeHistory('stock', 0, 'New QR-code created');
```
HaGuesto: This would mean that you write into the history table that something happened in the stock table and that the id of the data is 0.

PR #275, opened 2019-11-20
Added table list of foreign keys to listdelete

naphets123:


HaGuesto:
[Implement FK check for archiving #large](https://trello.com/c/k2q4w4OK/334-implement-fk-check-for-archiving-large)

PR #276, opened 2019-11-21
Speed up select all checkbox

HaGuesto:
The bottleneck was that every check of every line created an event to change the styling of the row.
This styling change is now called directly without an event.

If we want to speed up further we can replace the jquery lines with Vanilla js.

There are still open problems. The request to the server takes almost 30s to be resolved.

### DIFF
--- a/assets/js/magic.js
+++ b/assets/js/magic.js
@@ -147,13 +147,13 @@ $(function() {
             .find(".item-select:visible") //if find ever proves to be too slow, we could replace it with getelementbyclass from vanilla js which works 6 times faster.
             .prop("checked", is_checked)
             .closest("tr")
-            .toggleClass("selected")
+            .toggleClass("selected", is_checked)
             .promise()
             .done(function() { //only trigger the event change of a ".item-select" once
                 parent
                     .find(".item-select:visible:first")
                     .closest("tr")
-                    .toggleClass("selected");
+                    .toggleClass("selected", !is_checked);
                 parent.find(".item-select:visible:first").trigger("change");
             });
     });
--- a/assets/js/magic.js
+++ b/assets/js/magic.js
@@ -141,22 +141,33 @@ $(function() {
 
     // group select
     $(".group-select").change(function() {
-        var parent = $(this).closest(".table-parent");
-        var is_checked = $(this).is(":checked");
+        // Faster way of doing it
+        // var parent = $.makeArray(
+        //     $(".group-select").closest(".table-parent")
+        // )[0];
+        // var is_checked = $(this).is(":checked");
+        // var checkboxes = parent.getElementsByClassName("item-select");
+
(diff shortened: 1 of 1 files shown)


assets/js/magic.js
```diff
@@ -141,17 +141,32 @@ $(function() {
 
     // group select
     $(".group-select").change(function() {
```
HaGuesto: This part is the only thing I changed. The rest is formatting.

HaGuesto:
[Performance issue at select all for large amounts of data](https://trello.com/c/AJgc1Qe6/385-performance-issue-at-select-all-for-large-amounts-of-data)

assets/js/magic.js
```diff
@@ -141,17 +141,32 @@ $(function() {
...
+        // var parent = $.makeArray(
+        //     $(".group-select").closest(".table-parent")
+        // )[0];
+        // var is_checked = $(this).is(":checked");
+        // var checkboxes = parent.getElementsByClassName("item-select");
+
+        // for (var checkbox of checkboxes) {
+        //     checkbox.checked = is_checked;
+        // slected style of corresponding row needs to be selected 
+        // }
```
jamescrowley: Is this supposed to be here? :D

HaGuesto: nope

assets/js/magic.js
```diff
@@ -141,28 +141,22 @@ $(function() {
...
-                .prop("checked", false)
-                .trigger("change");
-        }
+        var parent = $(".group-select").closest(".table-parent");
+        var is_checked = $(this).is(":checked");
+        parent
+            .find(".item-select:visible") //if find ever proves to be too slow, we could replace it with getelementbyclass from vanilla js which works 6 times faster.
+            .prop("checked", is_checked)
+            .closest("tr")
+            .toggleClass("selected")
```
DurkoMatko: Hmm..I think if you have some checkboxes manually checked before clicking "Select All", this "blind" toggle would remove "selected" class from those rows...so the checkbox will be set correctly, but the row will not have a "selected" class. Explicitly setting the class rather than toggling should solve this.

DurkoMatko: Yupp I was able to create the described scenario:
1) select some rows manually
2) "Select all" checkbox
3) Those manually selected rows now lost the "selected" class:
image

4) OW! This actually then leads to an issue that when I "unselect all" and, they get "selected" class back - and therefore **altho there's no checkbox checked, the toolbar with list actions is visible**..cuz there are "selected" rows....the same happens vice-versa...when I check those rows, the class gets toggled (therefore removed)...now I have **checked & unselected rows and no toolbar** :D
image

image

HaGuesto: should be solved

jamescrowley:
:tada: apologies for the lack of follow-up on my part

PR #278, opened 2019-11-24
Add toast notifications to mobile (former "Replace nofilter flags with safer solution")

DurkoMatko:
**change**: replace all mobile notifications (DOM elements from smarty templates) by noty toast notifications (showing smarty DOM notification is still possible by passing message GET parameter)
**change**: adjusted cypress moibile tests to expect toast notifications

### DIFF
--- a/assets/js/mobile.js
+++ b/assets/js/mobile.js
@@ -24,3 +24,15 @@ function updateSizes(size){
 		$('#field_size_id').prepend('<option selected value="">Select a size</option>');
 	}
 }
+
+function saveBoxNotification(boxId, boxNumber, itemsCount, product, boxLocation){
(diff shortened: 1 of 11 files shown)


DurkoMatko:
[Not escape html in savebox on mobile](https://trello.com/c/J44LwGJ6/383-not-escape-html-in-savebox-on-mobile)

DurkoMatko:
[Not escape html in barcode.php](https://trello.com/c/pydMp0zb/382-not-escape-html-in-barcodephp)

jamescrowley:
@DurkoMatko how do you feel about this having tried the approach? I'm a little concerned I've nudged you in a direction that doesn't seem that nice now that I look at it, because of the number of random parameters you have to scatter all over the place.

DurkoMatko:
@jamescrowley It felt like making something work what's not supposed to be there 😄  We can do notifications by creating templates for notifications if this is something we don't want. And potentially hide/remove those DOM elements by some JS timer, so they act bit more like notifications. It's a pity that I've spent time on this but yeah, whatever..If it's not good, it's not good :D

HaGuesto:
closing due to redesign of corresponding task https://trello.com/c/PcDxAcMq

PR #279, opened 2019-11-27
User menu tests

cohenlea:


cohenlea:
[Cypress: User Mgmt. 2.7+, and select 2.8 tests #medium](https://trello.com/c/TFO51J17/343-cypress-user-mgmt-27-and-select-28-tests-medium)

cypress/integration/2_8_ExpiredUsers.js
```diff
@@ -0,0 +1,142 @@
...
+
+    it("2_8_5 Deactivate expired user", () => {
+        checkUserCheckboxByName(EXPIRED_COORDINATOR_NAME);
+        clickOnElementByTypeAndTestId("button", "list-delete-button");
+        checkForElementByText("h3", ARE_YOU_SURE_POPUP);
+
+        cy.server().route("POST", "/?action=cms_users_expired").as("expired");
+        clickOnElementByText("div.popover-content a", DEACTIVATE_BUTTON);
+
+        cy.wait("@expired").then(xhr => {
```
DurkoMatko: @jamescrowley I'm just not sure whether can we do this or is it possibly some kind of bad habit in UI/e2e tests? Basically, instead of clicking the grid tab, @cohenlea navigates to page of that tab. Doing this, she has the handle on the request and can wait till the page loads. I totally like it and if you approve of this, I'll do the same to hopefully fix our constantly failing 2.5 test :D I'm just not sure if it's "allowed"..cuz basically the navigation beween tabs by clicking is not tested in this case.

jamescrowley: 👍 if it's necessary to get our tests passing reliably, I'd take it right now!

But from what I can see, it's not so much about the tab navigation (which could still be done by clicking on it, I suspect?), but the 'hook' waiting on the POST submission. Is that right @cohenlea ?

There *might* be an argument to define all these 'intercepted' routes at the top of the test to make it clear they're not performing actions themselves. But, 👍

DurkoMatko: I thought/ still think, the POST submission is doing what the tab click would do - changing the page. Cuz every tab is basically it's own page, it's not that we have some JS driven grid which has several tabs to it. 

E.g the reason why 2.5 is failing is cuz we click on the tab and then want to check a checkbox in the new tab...but cypress doesn't wait for the new page(tab) to load because it finds the checkbox in the old tab already...then tab changes and there's no checkbox checked.

@cohenlea is preventing this by changing the tab by POST request. That way she gets the hook and can wait for the request to finish and be sure the correct tab is open (before she continues with assertions etc)

cohenlea: Hey @DurkoMatko and @jamescrowley ! Thanks for the review. For some reason I didn't get notified by github so I'm just seeing the comments now.

I'll move the cy.server() line into a separate function and add a comment explaining what it's doing, but basically, I'm just listening for that request to happen (I'm not clicking or initiating anything) and then checking the response code/message. I chose this strategy because I found the timing to be too unreliable for cypress to try to check for the success popup to appear (since the request has to finish first, but it's not actually changing tabs for this action).

DurkoMatko: Looking at it now, I got it all wrong @james :D it indeed isn't about navigation at all :D

PR #280, opened 2019-12-02
Rework dailyroutine

HaGuesto:
I finally had time to take a look at the dailyroutine and clean up.

### DIFF
--- a/dailyroutine.php
+++ b/dailyroutine.php
@@ -44,22 +44,28 @@
 
     // people that have not been active for a longer time will be deleted(Changed to deactivated in visible text, variables remain under the name deleted, as does the databasse)
     // the amount of days of inactivity is set in the camp table
-    $result = db_query('SELECT p.id, p.lastname, p.created, p.modified, c.delete_inactive_users AS treshold FROM people AS p LEFT OUTER JOIN camps AS c ON c.id = p.camp_id WHERE NOT p.deleted AND p.parent_id IS NULL');
+    $result = db_query('
+        SELECT p.id, p.lastname, p.created, p.modified, c.delete_inactive_users AS treshold 
+        FROM people AS p 
+        LEFT OUTER JOIN camps AS c ON c.id = p.camp_id 
+        WHERE (NOT p.deleted OR p.deleted IS NULL) AND p.parent_id IS NULL');
     while ($row = db_fetch($result)) {
-        /*
-        if(strtotime($row['lastfoodtransaction'])>strtotime($row['lastransaction'])) {$row['lastransaction'] = $row['lastfoodtransaction'];}
-        var_export($row['lastfoodtransaction']);
-        $row['touch'] = (strtotime($row['lasttransaction'])>strtotime($row['modified'])?$row['lasttransaction']:$row['modified']);
-        if(!$row['touch']) $row['touch'] = $row['created'];
-         */
-        $row['touch'] = db_value('SELECT GREATEST(COALESCE((SELECT transaction_date 
-					FROM transactions AS t 
-					WHERE t.people_id = people.id AND people.parent_id IS NULL AND product_id IS NOT NULL 
-					ORDER BY transaction_date DESC LIMIT 1),0), 
-					COALESCE(people.modified,0),COALESCE(people.created,0))
-				FROM people
-				WHERE id = :id', ['id' => $row['id']]);
-
+        $row['touch'] = db_value('
+            SELECT GREATEST(COALESCE((
+                SELECT transaction_date 
+				FROM transactions AS t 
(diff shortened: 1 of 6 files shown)


HaGuesto:
[Dailyroutine can not perform Drop capping routine](https://trello.com/c/Te2xYxHu/57-dailyroutine-can-not-perform-drop-capping-routine)

dailyroutine.php
```diff
@@ -5,70 +5,69 @@
...
-    transactions AS t,
-    camps AS c
-WHERE
-    t.people_id = p.id AND
-    NOT p.deleted AND
-    c.id = p.camp_id AND
-    p.parent_id IS NULL
-GROUP BY p.id
+    //--------- capping of tokens
+    $result = db_query('
```
jamescrowley: Just for my info, why is capping necessary in this way, vs just ensuring when we distribute they don't get too many?

dailyroutine.php
```diff
@@ -5,70 +5,69 @@
...
-        $j = 0;
-
-        $result2 = db_query('SELECT id FROM people WHERE NOT deleted AND parent_id = '.$row['id'].' ORDER BY DATE_FORMAT(FROM_DAYS(DATEDIFF(NOW(), date_of_birth)), "%Y")+0 DESC');
-        while ($row2 = db_fetch($result2)) {
-            ++$j;
-            db_query('UPDATE people SET seq = '.$j.' WHERE id = '.$row2['id']);
+        if ($row['drops'] > $row['maxdrops']) {
+            db_query(
+                '
+                INSERT INTO transactions (people_id, description, drops, transaction_date)
```
jamescrowley: we're not actually capping anything here, right? just logging that it would have been capped?

db/migrations/20191202131340_adult_age_in_camps_table_not_null.php
```diff
@@ -0,0 +1,15 @@
...
+use Phinx\Migration\AbstractMigration;
+
+class AdultAgeInCampsTableNotNull extends AbstractMigration
+{
+    public function change()
+    {
+        $this->table('camps')
+            ->changeColumn('adult_age', 'integer', [
+                'null' => false,
+                'default' => 15,
```
jamescrowley: Have you looked at the prod data? Do we need to do any migration or is everything non-null already?

dailyroutine.php
HaGuesto: I have no arguement for it. It has just always been this way. Should we change it?

dailyroutine.php
HaGuesto: No we are. it adds a capping transaction to the transaction table with a an amount of tokens passed to equalize.

db/migrations/20191202131340_adult_age_in_camps_table_not_null.php
HaGuesto: no migration is required

dailyroutine.php
jamescrowley: Hmm, I can see this is happening in give2all as well. So not sure why it's necessary at all?

dailyroutine.php
jamescrowley: Gotcha - mistook transactions for history table 🤦‍♂

dailyroutine.php
```diff
@@ -82,21 +81,17 @@
         }
     }
 
-    // cleaning up the database in case of errors
-    // delete children without a parent
-    db_query('UPDATE people p1 LEFT JOIN people p2 ON p1.parent_id = p2.id SET p1.deleted=NOW() WHERE p2.ID IS NULL AND p1.parent_id IS NOT NULL');
```
jamescrowley: Not necessary due to FKs now?

HaGuesto: exactly

dailyroutine.php
HaGuesto: I would propose merging it and creating a ticket to move this functionality to give2all. The capping is used by Skaramangas in Drop in the Ocean.

jamescrowley: Sorry, to be clear - it already happens in give2all already (https://github.com/boxwise/dropapp/blob/master/include/give2all.php#L49). Given this has been commented out for so long - have there actually been any side effects to this? Just wondering whether it's safer to actually leave it commented out.

HaGuesto: you are right. Thanks for checking and sorry that I was not aware of it. I'll comment it out.

jamescrowley:
@HaGuesto I took the liberty of deleting the commented out code. Don't need to leave it in, that's what source control is for :grin: Also removed the include at the top of the file, as that's included through the single entry point we have now anyway.

PR #281, opened 2019-12-02
Disable Stock FK check for deletion of products

HaGuesto:


HaGuesto:
[Products cannot be deleted due to FK check.](https://trello.com/c/qNIeAl8n/395-products-cannot-be-deleted-due-to-fk-check)

PR #283, opened 2019-12-02
Remove Faker library from Cypress seed

HaGuesto:
We discussed this in slack.

### DIFF
--- a/db/seeds/Cypress.php
+++ b/db/seeds/Cypress.php
@@ -13,12 +13,6 @@ class Cypress extends AbstractSeed
 
     public function run()
     {
-        // Faker library
-        // https://github.com/fzaninotto/Faker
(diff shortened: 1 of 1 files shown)


HaGuesto:
[Remove faker library from Cypress seed](https://trello.com/c/QzwL0gu2/401-remove-faker-library-from-cypress-seed)

HaGuesto:
The test should now actually be less flaky :)

DurkoMatko:
What do you mean by "should"? Were they for you locally? Or you mean "should" like that we think it could help....Cuz I think we talked that removing these data would speed up the tests and _potentially_ make the tests less flaky. But for me locally some still failed even with new seed..but at least they fail faster now :laughing:  So I merged it....but it didn't solve the issue that 2.5 and 5.2 fail often...

PR #284, opened 2019-12-02
Speed up export handling in magic.js

HaGuesto:
The export of large data is now down to a delay of a few seconds only and in the acceptable range.

The reason for the time delay was the creation of a collection called `allTargets` which is not used for exporting and the unselecting of all items in the list.

### DIFF
--- a/assets/js/magic.js
+++ b/assets/js/magic.js
@@ -149,7 +149,8 @@ $(function() {
             .closest("tr")
             .toggleClass("selected", is_checked)
             .promise()
-            .done(function() { //only trigger the event change of a ".item-select" once
+            .done(function() {
+                //only trigger the event change of a ".item-select" once
                 parent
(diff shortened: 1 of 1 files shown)


HaGuesto:
[Performance issue when exporting boxes - page becomes unresponsive. #medium](https://trello.com/c/GTKFBAjD/317-performance-issue-when-exporting-boxes-page-becomes-unresponsive-medium)

HaGuesto:
@DurkoMatko @jamescrowley When you have the time, it would be nice to get a review for this.

DurkoMatko:
@HaGuesto ok I've compared the performance even with just 350 boxes and it indeed feels bit more swift even with such small numbers! Nice :)

HaGuesto:
@DurkoMatko Thanks!

PR #286, opened 2019-12-04
Changed to include only non-deleted boxes in visible locations

naphets123:


HaGuesto:
[show only active boxes in the product menu. #tiny](https://trello.com/c/nuXjq1bi/328-show-only-active-boxes-in-the-product-menu-tiny)

PR #287, opened 2019-12-07
Don't escape select template options

DurkoMatko:
**change:** don't escape select options
**add:** parentheses around "beta" in needed items and fancy graphs

Parentheses were needed cuz in the select, it looked all just like inline text (the <sup> styling doesn't do anything there..). Looks like this now:

image


DurkoMatko:
[html escaping of beta functions in select fields](https://trello.com/c/KBdWzIGE/398-html-escaping-of-beta-functions-in-select-fields)

jamescrowley:
@DurkoMatko perhaps could just switch to putting beta in brackets and avoid the 'nofilter'? Otherwise in theory we'd need to check the other possible routes into displaying drop down data to check if those are safe to not be encoded too, given cms_form_select.tpl is global

DurkoMatko:
@jamescrowley Hmmm right...so I've removed the nofilter and also removed the sup HTML tag from both beta functions.

HaGuesto:
@jamescrowley it does not show then beta in superscript. Is that what you wanted?

jamescrowley:
@HaGuesto it's not ideal, but as discussed - was the 'simplest' way to preserve the HTML encoding in the global templates/cms_form_select.tpl file

PR #294, opened 2019-12-13
add selected attribute to form option if only one option for field

cohenlea:
hope this isn't a terrible idea...


cohenlea:
[Auto-select option if there is only a single one(e.g. boxes size: onesize) #tiny](https://trello.com/c/iMJWtaJU/388-auto-select-option-if-there-is-only-a-single-oneeg-boxes-size-onesize-tiny)

library/ajax/getsizes.php
```diff
@@ -3,11 +3,13 @@
     $success = true;
     $product_id = $_POST['product_id'];
     $size_id = intval($_GET['size']);
+    $one_size_id = 68;
```
HaGuesto: It is not a good idea to hard_code db values here. Besides, there is also the sizegroup `Mixed sizes` which also just has one option.

library/ajax/getsizes.php
```diff
@@ -3,11 +3,13 @@
...
 
     $result = db_query('SELECT s.* FROM sizes AS s, sizegroup AS sg, products AS p WHERE s.sizegroup_id = sg.id AND p.id = :id AND p.sizegroup_id = sg.id ORDER BY s.seq', ['id' => $product_id]);
```
HaGuesto: could you replace this with a `db_array` command, so that it fetches all results at once. In the while loop instead of fetching one by one you can then check the length of the array and add the selected class if the length of the array is one.

templates/cms_form_select.tpl
```diff
@@ -13,6 +13,7 @@
 			<option></option>
 			{foreach $element['options'] as $option}
 				<option 
+					{if count($element['options']) == 1 && $element['field'] != "product_id"}selected {/if}
```
HaGuesto: Why do you need this line? The `getSizes()` ajax call replaces all options. I do not think it is needed.

cohenlea: this is for other fields such as warehouse locations

PR #300, opened 2019-12-15
Add qr_id to mobile box formhandler only if POST['qr_id'] exists

HaGuesto:


HaGuesto:
[edit of comments on mobile do not work.](https://trello.com/c/gSGJPDre/409-edit-of-comments-on-mobile-do-not-work)

PR #301, opened 2019-12-17
check for open class on confirm popup

cohenlea:
Instead of checking for the "Are you Sure?" popup in the DOM, check for the "open" class in the "Activate" button.

### DIFF
--- a/cypress/integration/2_8_ExpiredUsers.js
+++ b/cypress/integration/2_8_ExpiredUsers.js
@@ -39,6 +39,14 @@ describe("2_8_ExpiredUsers_Test", () => {
         cy.get(type + "[data-testid = '" + testId + "']").should("be.visible");
     }
 
+    function checkClassByTypeAndTestId(type, testId, _class, hasClass) {
+        if (hasClass == true) {
+            cy.get(type + "[data-testid = '" + testId + "']").should("have.class", _class);
+        } else {
+            cy.get(type + "[data-testid = '" + testId + "']").should("not.have.class", _class);
+        }
+    }
+
     function clickOnElementByTypeAndTestId(type, testId) {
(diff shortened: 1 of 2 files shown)


cohenlea:
[Cypress: failing tests 2_7 2_8](https://trello.com/c/AmjqQZrN/420-cypress-failing-tests-27-28)

cypress/integration/2_8_ExpiredUsers.js
```diff
@@ -39,6 +39,14 @@ describe("2_8_ExpiredUsers_Test", () => {
         cy.get(type + "[data-testid = '" + testId + "']").should("be.visible");
     }
 
+    function checkClassByTypeAndTestId(type, testId, _class, hasClass) {
```
HaGuesto: It might be cleaner to define this function cypress wide in `cypress/support`. Is there a reason why you did not do it? @cohenlea

cohenlea: I can, but is there a reason to move this specific one vs the other 10 or so functions defined in the test itself? There are many functions we repeat in each test file (such as find or click by text or test id) so maybe would be good to have a big cleanup and move them all to the a support file and then make that the best practice for everyone moving forward. Otherwise not sure why this one stood out to you to be separated? @HaGuesto

HaGuesto: Yes. So, I have not reviewed many test before. It just stood out in this PR that you are defining the same function twice.

HaGuesto: I will add a ticket.

cohenlea: Sounds good @HaGuesto !

PR #302, opened 2019-12-17
Track QR Box connection in History table

HaGuesto:


HaGuesto:
[Track the association of a qr code to a box when a box is created on mobile](https://trello.com/c/YV3JC8oE/423-track-the-association-of-a-qr-code-to-a-box-when-a-box-is-created-on-mobile)

PR #303, opened 2019-12-17
Replace trigger_error with Exceptions on desktop environment

HaGuesto:


HaGuesto:
[check the handling of all trigger_error calls](https://trello.com/c/eVdPUAjI/414-check-the-handling-of-all-triggererror-calls)

PR #304, opened 2019-12-18
Logging all Box creation to sentry

HaGuesto:


HaGuesto:
[logging Box creation in the form handler with reporting to sentry](https://trello.com/c/pnrkCXQm/426-logging-box-creation-in-the-form-handler-with-reporting-to-sentry)

PR #309, opened 2019-12-23
Fix mobile checkout cart editing

DurkoMatko:
**change**: interactable elements in shopping row have higher z-index than following title


DurkoMatko:
[Checkout shopping cart is not editable on mobile screen sizes #medium](https://trello.com/c/xMkPrWhi/422-checkout-shopping-cart-is-not-editable-on-mobile-screen-sizes-medium)

PR #314, opened 2020-01-08
Rework ajax errors

naphets123:


HaGuesto:
[Rework ajax errors on desktop screen #large](https://trello.com/c/JLdkcUCd/133-rework-ajax-errors-on-desktop-screen-large)

assets/js/custom.js
```diff
@@ -151,6 +142,47 @@ function offsetAnchor() {
...
+            }
+            
+
+        if (result.action) {
+            eval(result.action);
+            }
+
+            
+    }
+    else
```
HaGuesto: This if/else does not make sense and causes many tests to fail.
How you have written it it only posts messages if there was no error on the server-side. However, there are cases like login errors were other error messages should be shown.

assets/js/magic.js
```diff
@@ -422,9 +422,10 @@ $(function() {
                         dataType: "json",
                         success: function(checkresult) {
                             if (checkresult.success) {
-                                AjaxFormSubmit(form);
```
HaGuesto: @naphets123 Why did you move this line?

naphets123: i put it back

assets/js/custom.js
naphets123: Ok, I removed the whole conditional on succes, it was actually just used for everything except message/redirect/action

HaGuesto: However, you removed to much now. 
You neglected the case 3) now from the trello card.
>Distinguish between three errors:
>1) Ajax request does not get a response from the server. The notification should be
>"Cannot connect to Boxwise - please check your Internet connection."
>2) The response is a json looking like {'success': false, 'message':'Something something', ....}
>If a message is specified please show it to the user.
>3) A response is received, but it is not in the format of 2). The notification should be
>"Something went wrong - please inform your coordinator."

HaGuesto: I will tag the line where you should put it.

assets/js/custom.js
```diff
@@ -151,6 +142,38 @@ function offsetAnchor() {
         window.scrollTo(window.scrollX, window.scrollY - 100);
     }
 }
+
+function AjaxCheckSuccess(result){
+        if (result.message){
```
HaGuesto: Add here something like
```
if (result.success does not exist) {
            var n = noty({
                text: "Something went wrong - please inform your coordinator.",
                type: "error"
                });
        }

File: assets/js/custom.js [hunk 4]
```diff
@@ -151,6 +142,43 @@ function offsetAnchor() {
...
+            var n = noty({
+                text: result.message,
+                type: result.success ? "success" : "error"
+                });
+        }
+        if (result.redirect){
```
HaGuesto: This if clause does not make sense. It should look like
```
                if (result.redirect) {
                    if (result.message) {
                        setTimeout(function() {
                            execReload(result.redirect);
                        }, 1500);
                    } else {
                        execReload(result.redirect);
                    }
                }
```

HaGuesto: you left out the second if clause. The redirect is timed out for the time a notification needs to be shown.

assets/js/custom.js
```diff
@@ -378,19 +388,10 @@ function getSizes() {
...
-            }
+            AjaxCheckSuccess(result);
         },
         error: function(result) {
-            var n = noty({
-                text:
-                    "Something went wrong, maybe the internet connection is a bit choppy",
-                type: "error"
-            });
+            AjaxError(result);
```
jamescrowley: Just a little note that you can simplify

```
error: function(result) {
     AjaxError(result);
},
```

to

`error: AjaxError,`

and so on - ie just pointing to the function definition :)

assets/js/custom.js
naphets123: Ok I changed this and applied the shorter syntax james proposed below. I think this fixes the issue you saw with the login tests

assets/js/custom.js
```diff
@@ -378,20 +386,10 @@ function getSizes() {
...
+            AjaxCheckSuccess(result);
         },
-        error: function(result) {
-            var n = noty({
-                text:
-                    "Something went wrong, maybe the internet connection is a bit choppy",
-                type: "error"
-            });
-        }
+        error: AjaxError(result)
```
HaGuesto: This throws a timeout error in Test 3_2.

The browser console shows a Reference Error.
```
custom.js?v=7:393 Uncaught ReferenceError: result is not defined
    at getSizes (custom.js?v=7:393)
    at HTMLSelectElement.onchange (?action=stock_edit&origin=stock:201)
    at Object.trigger (minified.js:1)
    at HTMLSelectElement.<anonymous> (minified.js:1)
    at Function.each (minified.js:1)
    at r.fn.init.each (minified.js:1)
    at r.fn.init.trigger (minified.js:1)
    at d.triggerChange (minified.js:348)
    at d.onSelect (minified.js:348)
    at d.selectHighlighted (minified.js:348)
```

HaGuesto: @naphets123 I figured the Error out and fixed it. Ready to merge now.

PR #316, opened 2020-01-10
Working attack that uses coordinator account to change coordinator an…

naphets123:
Don't merge this. Just an example of how to create manipulated ajax-requests to change credentials, in this case the coordinator account is used to make itself and another ordinary user into admins.


HaGuesto:
@naphets123 close if it is not needed. Please link this PR in the Google Doc.

aerinsol:
Stephan, I think we should talk about this and medium-to-long-term plans on
how to handle this issue. PM me on Slack.

PR #317, opened 2020-01-12
remove counter from mobile findbox input field

cohenlea:


cohenlea:
[Mobile: Search Box ID should not have a counter #tiny](https://trello.com/c/vni1GKrH/432-mobile-search-box-id-should-not-have-a-counter-tiny)

PR #318, opened 2020-01-13
Edit usergroups

naphets123:


naphets123:
[Edit usergroups](https://trello.com/c/4aq18G7k/443-edit-usergroups)

naphets123:
This goes a little further than just the original issue in the trello card, but i felt like these issues are very closely related and should be handled at once. Now it should be unable to access user via manipulation of the url without valid access. 
Also it should now be impossible to edit cms_users or cms_usergroups by sending a custom-ajax call bypassing the form. 
Additionally cms_usergroups of other camps, which were visible by changing the id in the url should now lead to a direct redirect to cms_usergroups. This is the same way this was already handled for cms_users/cms_users_edit.

aerinsol:
Nice work Stephan! Yes, this needs to be handled, but you weren't on the
dev call this weekend so I wasn't able to make an announcement to you
properly with the team. One of the first orders of this year is we will
build a proper back-end API to handle authentication and authorization on
the new tech stack. So probably you and Lea will be working on that as soon
as we finalize the tech stack decision and manage to get it up and running.

include/cms_usergroups_edit.php
```diff
@@ -5,9 +5,18 @@
...
-        if ($_SESSION['user']['is_admin'] || ($_SESSION['usergroup']['userlevel'] > db_value('SELECT level FROM cms_usergroups_levels WHERE id = :id', ['id' => $_POST['userlevel']]))) {
+        $postedgroup = db_row('
+			SELECT ug.organisation_id, ugl.level AS userlevel
+			FROM cms_usergroups AS ug
+			LEFT JOIN cms_usergroups_levels AS ugl ON ugl.id=ug.userlevel
+            WHERE ug.id = :id AND (NOT ug.deleted OR ug.deleted IS NULL)', ['id' => $_POST['id']]);
+        $is_admin = $_SESSION['user']['is_admin'];
+        $allowed_group_level = ($_SESSION['usergroup']['userlevel'] > $postedgroup['userlevel']);
+        $allowed_organisation = ($_SESSION['usergroup']['organisation_id'] == $postedgroup['organisation_id']);
+        $allowed_new_level = ($_SESSION['usergroup']['userlevel'] > db_value('SELECT level FROM cms_usergroups_levels WHERE id = :id', ['id' => $_POST['userlevel']]));
```
HaGuesto: @naphets123 Is this not a duplicate of `$allowed_group_level` two lines above?

include/cms_usergroups_edit.php
```diff
@@ -22,6 +31,14 @@
     }
 
     $data = db_row('SELECT * FROM '.$table.' WHERE id = :id', ['id' => $id]);
+    $requestedgroup = db_row('
+		SELECT ug.organisation_id, ugl.level AS userlevel
+		FROM cms_usergroups AS ug
+		LEFT OUTER JOIN cms_usergroups_levels AS ugl ON ugl.id=ug.userlevel
+		WHERE ug.id = :id AND (NOT ug.deleted OR ug.deleted IS NULL)', ['id' => $data['id']]);
+    if (!$_SESSION['user']['is_admin'] && ($data && (($_SESSION['organisation']['id'] != $requestedgroup['organisation_id']) || ($_SESSION['usergroup']['userlevel'] <= $requestedgroup['userlevel'])))) {
+        redirect('?action='.$_GET['origin']);
```
HaGuesto: Please replace the redirect with an exception saying "You do not have access to this user group!"

include/cms_usergroups_edit.php
naphets123: No, one is the groups i am allowed to modify the other one groups i can modify them into

HaGuesto: yes you are right.

include/cms_usergroups_edit.php
```diff
@@ -5,9 +5,25 @@
 
 if ($_SESSION['user']['is_admin'] || $_SESSION['usergroup']['userlevel'] > db_value('SELECT MIN(level) FROM cms_usergroups_levels')) {
     if ($_POST) {
-        if ($_SESSION['user']['is_admin'] || ($_SESSION['usergroup']['userlevel'] > db_value('SELECT level FROM cms_usergroups_levels WHERE id = :id', ['id' => $_POST['userlevel']]))) {
-            $_POST['organisation_id'] = $_SESSION['organisation']['id'];
+        // Distinguish between create and update of a user group
+        if ($_POST['id']) {
```
HaGuesto: @naphets123 you missed distinguishing between creating and updating a user group.

PR #319, opened 2020-01-14
Prevent access to non-valid people ids in the people menu by POST req…

naphets123:
…uest


naphets123:
[access rights in POST request to people.php and people_deactivated.php is not valid correctly](https://trello.com/c/3aIlBHMT/411-access-rights-in-post-request-to-peoplephp-and-peopledeactivatedphp-is-not-valid-correctly)

PR #327, opened 2020-02-03
Add Children by age sizegroup with two year interval

HaGuesto:


HaGuesto:
[Create new 2year interval children's size range for R4R#small](https://trello.com/c/9AhS8hFZ/453-create-new-2year-interval-childrens-size-range-for-r4rsmall)

PR #329, opened 2020-02-06
Outgoing warehouse locations for users

HaGuesto:


HaGuesto:
[Update 'Warehouse' Menu Function to make it safe to expose to regular users in through usergroup permissions](https://trello.com/c/Vmap6T6v/433-update-warehouse-menu-function-to-make-it-safe-to-expose-to-regular-users-in-through-usergroup-permissions)

aerinsol:
Why are we adding an is_scrap flag?

HaGuesto:
That was quick. The `is_scrap` field is needed since we have outgoing locations that are called `SCRAP`. Boxes which contain damaged clothes are moved there.

When I updated the query, I realized that the user was still able to edit these SCRAP locations. Therefore, I added a flag to mark SCRAP locations to remove the access of the user to these locations.

aerinsol:
Haha, it's because I'm teaching my Mom how to use Gmail.

Your logic sounds good.

PR #334, opened 2020-02-13
Fix search in warehouse locations

HaGuesto:


HaGuesto:
[Search in warehouse locations does not work](https://trello.com/c/lhEfwYbs/470-search-in-warehouse-locations-does-not-work)

PR #335, opened 2020-02-13
Fix data access control of beneficiary for move

HaGuesto:


HaGuesto:
[Drag and drop of beneficiaries throws error.](https://trello.com/c/jmQUVwKP/471-drag-and-drop-of-beneficiaries-throws-error)

PR #338, opened 2020-02-15
Bulk extend validity date button

DurkoMatko:
add: prolong list button for expired/active users


library/lib/list.php
```diff
@@ -208,6 +208,34 @@ function listUnDeleteAction($table, $id, $count = 0, $recursive = false)
     return $count;
 }
 
+function listProlong($table, $ids)
+{
+    global $translate;
+
+    $hasExpireDate = db_fieldexists($table, 'valid_lastday');
+    foreach ($ids as $id) {
+        if ($hasExpireDate) {
```
DurkoMatko: This condition is executed for every ID. It'd perform better if it was vice-versa. It's the case for several functions in this `list.php` file so once this PR is merged, I'll refactor the file.

DurkoMatko:
[Bulk Extend Validity Date for Users in User Management #medium](https://trello.com/c/fJwEJNRF/372-bulk-extend-validity-date-for-users-in-user-management-medium)

HaGuesto:
@DurkoMatko please check the updates in the trello card.

db/seeds/MinimalDb.php
```diff
@@ -482,6 +482,8 @@ public function run()
 			(901,13,'text','cms_list_undelete',NULL,'Herstellen','Recover','',0,NULL,NULL,NULL,NULL,'0000-00-00 00:00:00'),
 			(902,13,'text','cms_list_undeleteerror',NULL,'De geselecteerde onderdelen (of een deel ervan) konden niet worden hersteld.','The selected items (or some of them) could not be recovered.','',0,NULL,NULL,NULL,NULL,'0000-00-00 00:00:00'),
 			(903,13,'text','cms_list_undeletesuccess',NULL,'Onderdeel hersteld','Item recovered','',0,NULL,NULL,NULL,NULL,'0000-00-00 00:00:00'),
+			(904,13,'text','cms_list_extendsuccess',NULL,'Vervaldatum item verlengd','Item expiration extended','',0,NULL,NULL,NULL,NULL,'0000-00-00 00:00:00'),
```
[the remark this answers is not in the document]
HaGuesto: like here just with insert instead of DROP
https://github.com/boxwise/dropapp/blob/00a779df0ab788645dbfe8c445c8f7d4b8e36869/db/migrations/20200313200406_remove_needed_items.php#L13

library/lib/list.php
HaGuesto: I see what you mean. However, the effect of it is purely esthetical. 
Since you set the variable `$hasExpiryDate` to lines before, the difference performance-wise are just if-else checks ho literally have no performance cost.

include/cms_users_expired.php
```diff
@@ -6,6 +6,7 @@
         require_once 'cms_users_handle_ajax_operations.php';
     } else {
         initlist();
+        addbutton('extend', 'Extend Access', ['icon' => 'fa-history', 'confirm' => true, 'testid' => 'extend-cms-user']);
```
HaGuesto: You need to add here the predefined options as 
`'options'=>['day'='One Day',....]`

include/cms_users.php
```diff
@@ -12,6 +12,7 @@
         addpagemenu('active', 'Active & Pending', ['link' => '?action=cms_users', 'active' => true, 'testid' => 'active_pending']);
         addpagemenu('expired', 'Expired', ['link' => '?action=cms_users_expired', 'testid' => 'expired']);
         addpagemenu('deactivated', 'Deactivated', ['link' => '?action=cms_users_deactivated', 'testid' => 'deactivated']);
+        addbutton('extendActive', 'Extend Access', ['icon' => 'fa-history', 'confirm' => true, 'testid' => 'extend-active-cms-user']);
```
HaGuesto: You need to add here the predefined options as 
`'options'=>['day'='One Day',....]`

include/cms_users_handle_ajax_operations.php
```diff
@@ -22,11 +22,17 @@
...
             if ($success) {
                 foreach ($ids as $id) {
-                    db_query('UPDATE cms_users SET email = SUBSTR(email, 1, LENGTH(email)-LENGTH(".deleted.")-LENGTH(id)) WHERE id = :id', ['id' => $id]);
+                    db_query('UPDATE '.$table.' SET email = SUBSTR(email, 1, LENGTH(email)-LENGTH(".deleted.")-LENGTH(id)) WHERE id = :id', ['id' => $id]);
                 }
             }
 
             break;
+        case 'extendActive':
+        case 'extend':
```
HaGuesto: Since the extend functionality is quite custom. I'm not sure if I would library functions, but since you already have done that you can leave it in lib.
You need to handle the different options though.

include/cms_users_expired.php
DurkoMatko: Wow nice

DurkoMatko:
As mentioned in Slack discussion with @HaGuesto, this one is review/merge-ready

library/lib/list.php
```diff
@@ -208,6 +208,58 @@ function listUnDeleteAction($table, $id, $count = 0, $recursive = false)
     return $count;
 }
 
+function listExtend($table, $ids, $period)
+{
+    global $translate;
+
+    $hasExpireDate = db_fieldexists($table, 'valid_lastday');
+    $updatedValue = [];
```
HaGuesto, line 216: I changed it so hat an array is returned. This is needed if multiple rows are extended at the same time.

library/lib/list.php
```diff
@@ -208,6 +208,58 @@ function listUnDeleteAction($table, $id, $count = 0, $recursive = false)
...
+    }
+
+    return [false, $translate['cms_list_extenderror'], false];
+}
+
+function listExtendAction($table, $id, $period)
+{
+    switch ($period) {
+        case 0:
+            $extendQuery = 'UPDATE '.$table.' SET valid_lastday = DATE_ADD(IF(valid_lastday AND NOT valid_lastday IS NULL AND valid_lastday > CURDATE(), valid_lastday, CURDATE()),  INTERVAL 1 WEEK) WHERE id = :id;';
```
HaGuesto, line 238: The new if statement solves two bugs:
1. If the `valid_lastday` is zero the extension should start from the current day.
2. if an account is already expired then the extension should also start from the current day. (From the user point of view I guess this makes more sense.)

PR #339, opened 2020-02-20
Unify filters

naphets123:


HaGuesto:
@naphets123 Let's close this PR for now since it is not needed atm. We can reopen it when we have troubles with filters again.

PR #350, opened 2020-03-13
replace lastname with surname

HaGuesto:


HaGuesto:
["Lastname" in Add Beneficiaries form should be "Surname" #tiny](https://trello.com/c/IqnWYVF6/480-lastname-in-add-beneficiaries-form-should-be-surname-tiny)

PR #359, opened 2020-05-02
Fix stock overview

HaGuesto:


HaGuesto:
[Stock Overview does not show Hygiene categories for Hermine #nextrelease](https://trello.com/c/Fb8aUNGc/509-stock-overview-does-not-show-hygiene-categories-for-hermine-nextrelease)

PR #360, opened 2020-05-02
Fix delete button of transactions in people_edit

HaGuesto:


HaGuesto:
[Deletion of transaction in beneficiary account does not work](https://trello.com/c/0UjIhCEd/514-deletion-of-transaction-in-beneficiary-account-does-not-work)

PR #365, opened 2020-05-30
Proposed Architecture Decision Process

HaGuesto:
@jamescrowley @aerinsol @boxwise/dev-member 
## Proposed Architecture Decision Process
To structure the decision-making process and to record all central architecture decisions in a central location (in each repo), I would like to propose the following process:

1. When you are facing an issue and you need to make an architecture decision, please create a copy of the `adr_template` file and fill it out.
2. The moment you are ready to make a proposal please create a Pull Request to master.
3. In the document you can set a deadline until when you need a decision.
4. Please request reviews from the people you want input. Remind them to review in slack if nothing happens.
5. Depending on the input, please go back to step 1. and build a new proposal. Best include the reviewer with the input in the new proposal.
6. If your proposal was approved, please merge the PR.

Let me know what you think.


pylipp:
Since I only joined recently, I can't say how the situation was like before. I like the suggestion though because it makes decisions quite visible for me _as developer_; as far as there no risk that decisions, or rather the decision finding processes, are not distributed between several platforms, like GitHub and Trello, and Slack.
Also we have to remember to maintain 'old' ADRs in case architecture changes but that hopefully comes naturally...

pylipp:
Are there any already existing decisions that you'd like to have an ADR for in hindsight?

jamescrowley:
@pylipp to answer your question re previous decisions, there's a Google Doc gives some picture of recent decisions made - I'll send it in Slack, as it also references some security issues that I don't want to highlight here :)

aerinsol:
Almost all of the stuff on the new app after Python React decision (which
was in concert with James) is a to-do for me to write up :)

PR #370, opened 2020-06-14
New UI theme (mobile)

DurkoMatko:
change: mobile CSS to match new theme


DurkoMatko:
[Update and unify CSS rules so behavior of mobile.php matches that of the new theme #small](https://trello.com/c/hM1b3UUM/558-update-and-unify-css-rules-so-behavior-of-mobilephp-matches-that-of-the-new-theme-small)

PR #375, opened 2020-07-02
Invite user

naphets123:


HaGuesto:
[Integrate PHP app with auth0 for identity management (authentication) only #php #extra-large](https://trello.com/c/8KEgl3nv/503-integrate-php-app-with-auth0-for-identity-management-authentication-only-php-extra-large)

PR #376, opened 2020-07-03
Fix filter custom outgoing

cohenlea:
We were not considering custom outgoing locations in the Boxes filter query.


cohenlea:
[Outgoing locations filter in Boxes menu shows only internal stock](https://trello.com/c/TPk0cpJW/571-outgoing-locations-filter-in-boxes-menu-shows-only-internal-stock)

PR #378, opened 2020-07-06
update logo

HaGuesto:


HaGuesto:
[BUG: QR Labels should use color version](https://trello.com/c/lPh8tXt7/575-bug-qr-labels-should-use-color-version)

PR #379, opened 2020-07-06
Drop Organisation Id from cms_users

HaGuesto:


HaGuesto:
[organisation_id in cms_users is null in most cases](https://trello.com/c/prILxpCs/574-organisationid-in-cmsusers-is-null-in-most-cases)

PR #381, opened 2020-07-06
Add Unique Index to CrossReference Tables

HaGuesto:


HaGuesto:
[Enforce composite primary keys on cross-reference tables](https://trello.com/c/B37cY4aN/573-enforce-composite-primary-keys-on-cross-reference-tables)

PR #384, opened 2020-08-08
Create user in Auth0 from php User Mgmt.

HaGuesto:


HaGuesto:
[Create user in Auth0 from #php](https://trello.com/c/Fr2mePV6/602-create-user-in-auth0-from-php)

PR #387, opened 2020-08-30
Move loading spinner from footer to header

HaGuesto:


HaGuesto:
[Loading Spinner is shown too late so that user can interact with page before it is fully loaded](https://trello.com/c/drLQGsqL/630-loading-spinner-is-shown-too-late-so-that-user-can-interact-with-page-before-it-is-fully-loaded)

PR #389, opened 2020-09-23
Fix people tags error messages

HaGuesto:


HaGuesto:
[Add error message for tags](https://trello.com/c/3Uao4d48/635-add-error-message-for-tags)

PR #391, opened 2020-10-05
Fix people search box with tags

HaGuesto:


HaGuesto:
[URGENT CHECK Manage Beneficiary search does not work most likely connected to tag filters. #nextrelease](https://trello.com/c/wN7HjAys/640-urgent-check-manage-beneficiary-search-does-not-work-most-likely-connected-to-tag-filters-nextrelease)

PR #394, opened 2020-10-25
Removing Bicycle and Workshop Features

HaGuesto:


HaGuesto:
[db clean up (Remove fields we do not use anymore)](https://trello.com/c/yZ2dbz61/655-db-clean-up-remove-fields-we-do-not-use-anymore)

aerinsol:
Wait wait wait, let's not remove the bike and borrow features yet. This may
be useful going forward.

Laundry and workshop, yes

HaGuesto:
@aerinsol At the moment I'm leaving in the library functionality which is still enabled for Drop in the Ocean. 

The bicycle and workshop stuff is not used anywhere and in our meeting we agreed to remove the database fields for bicycle. I would not like to leave code snippets regarding workshop and bicycle in there if we are removing the db fields because it is an easy source of bugs.

Let me know if I can merge or please give me a reason why you would like to keep the bicycle stuff besides the library functionality and I'm happy to keep it.

aerinsol:
Ah yes, I remember now. I'm hesitant only because rentals of bikes seems similar to rental of books. Can you take some screenshots or video of the bicycle functionality from staging? I don't think I've ever seen it, and I'd like to have it for reference.

aerinsol:
The reason I'd like to keep rental functionality is because from a base management or distributions / clothing bank point of view, there is often a need to "loan" out assets to people. We haven't seen this yet, but I'd like to keep the functionality so if the request comes up, we can at least give it to people to test, rather than coding it up from scratch.

HaGuesto:
Alright I won't merge this then. We should talk about this in our next meeting because library, bicycle and workshop are the same functionality about lending things out. Library is the most reliable. Bicycle and Workshop has additional features like banning someone from using bikes or the workshop. However, Bikes and Workshops are quite messily programmed with quite some hard-coded parts.

aerinsol:
Okay, sounds good. Please keep the PR since I mostly just want to see the
functionality and might want to merge this in later - if we can repurpose
library at some point that could be an option.

HaGuesto:
@aerinsol I enabled the Bicycle functionality for BoxAid on Lesvos on staging. Feel free to click around.

aerinsol:
I'm actually OK with merging this in now @HaGuesto

aerinsol:
As discussed, closing this due to age. This will go back on the backlog. https://trello.com/c/yZ2dbz61

PR #398, opened 2020-11-12
Add more tracing to people.php

HaGuesto:


HaGuesto:
[Add more tracing and limit GAE instances to diagnose Beneficiary performance](https://trello.com/c/VfYvEOYv/683-add-more-tracing-and-limit-gae-instances-to-diagnose-beneficiary-performance)

PR #399, opened 2020-11-26
Optimize People SQL Query

HaGuesto:
@aerinsol I will reply to the questions in the query here in the PR.

### DIFF
--- a/include/people.php
+++ b/include/people.php
@@ -91,7 +91,7 @@ Tracer::inSpan(
             $data = getlistdata('
             SELECT
                 people_filtered_with_tags.*,
-                SUM(CASE WHEN people_filtered_with_tags.level = 0 THEN transactions.drops ELSE 0 END) AS tokens,
+                IFNULL(SUM(CASE WHEN people_filtered_with_tags.level = 0 THEN transactions.drops ELSE 0 END),0) AS tokens,
                 MAX(transactions.transaction_date) AS last_activity
             FROM
                 (SELECT
@@ -152,7 +152,7 @@ Tracer::inSpan(
             LEFT JOIN
                 people AS parent ON people_filtered_with_tags.parent_id = parent.id
             LEFT JOIN
-                transactions ON transactions.people_id = people_filtered_with_tags.id AND transactions.product_id IS NOT NULL '.
+                transactions ON transactions.people_id = people_filtered_with_tags.id '.
             (
                 'approvalsigned' == $listconfig['filtervalue'] ? '
                 WHERE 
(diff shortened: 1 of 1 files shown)


include/people.php
```diff
@@ -91,7 +91,7 @@ function () use ($action, &$cmsmain) {
             $data = getlistdata('
             SELECT
                 people_filtered_with_tags.*,
-                IF(people_filtered_with_tags.parent_id,"",(SELECT SUM(drops) FROM transactions WHERE people_id = people_filtered_with_tags.id)) AS tokens,
+                SUM(CASE WHEN people_filtered_with_tags.level = 0 THEN transactions.drops ELSE 0 END) AS tokens,
```
HaGuesto, line 94: @aerinsol I removed the IFNULL bit since otherwise there would be a 0 in every line in the tokens column on the front-end.

include/people.php
```diff
@@ -113,13 +113,15 @@ function () use ($action, &$cmsmain) {
...
                         people.approvalsigned
                     FROM
-                        people
-                    LEFT JOIN
-                        people_tags AS people_tags_filter ON people_tags_filter.people_id = people.id 
-                    LEFT JOIN
-                        tags AS tags_filter ON tags_filter.id = people_tags_filter.tag_id AND tags_filter.deleted IS NULL AND tags_filter.camp_id = '.$_SESSION['camp']['id'].'
+                        people'.
+                    // Join tags here only if a tag filter is selected and only people with a certain tag should be returned
+                    ($listconfig['multiplefilter_selected'] ? '
```
HaGuesto, line 118: @aerinsol I removed quite a few WHERE clauses in line 124 and below since they are not important for the query where all beneficiaries are shown.
However, the redundant join of tags is needed if we want to filter people by tags. What I did now is just to make the JOINs optional (only if a filter for tags is supplied)

include/people.php
```diff
@@ -131,20 +133,26 @@ function () use ($action, &$cmsmain) {
...
-                    tags ON tags.id = people_tags.tag_id AND tags.deleted IS NULL AND tags.camp_id = '.$_SESSION['camp']['id'].'
+                    tags 
+                    ON tags.id = people_tags.tag_id 
+                    AND tags.deleted IS NULL 
+                    AND tags.camp_id = '.$_SESSION['camp']['id'].'
                 GROUP BY 
                     people_filtered.id
                 ) AS people_filtered_with_tags
             LEFT JOIN
                 people AS parent ON people_filtered_with_tags.parent_id = parent.id
```
HaGuesto, line 153: @aerinsol Unfortunately, this join is not only necessary for the ordering but also for the optional WHERE clause in line 150. 
This is the filter if you want to filter for all families (not single people) who have not signed the data privacy declaration.

There might be a better way to do this. Would be nice to get rid of this join.

HaGuesto:
[Performance of Manage Beneficiaries](https://trello.com/c/HdJOWVXg/661-performance-of-manage-beneficiaries)

include/people.php
aerinsol: @HaGuesto won't this still print a 0 on the front-end for children anyway?

include/people.php
aerinsol: @HaGuesto nice.

include/people.php
aerinsol: @HaGuesto which join specifically, is it this one, and if not, can you mention the line number?

```
            LEFT JOIN
                people AS parent ON people_filtered_with_tags.parent_id = parent.id
```
 
I don't remember making any changes to approval or data privacy logic.

include/people.php
HaGuesto: yes my mistake. u are right. solved it somewhere else and put your solution back in place.

include/people.php
HaGuesto: > @HaGuesto which join specifically, is it this one, and if not, can you mention the line number?
>             LEFT JOIN
>                 people AS parent ON people_filtered_with_tags.parent_id = parent.id
the one we are commenting on here. line 145. 

> I don't remember making any changes to approval or data privacy logic. 
That's right I removed line 150 and following when sending you the query because it included a filter which was useless to optimize the query to get all beneficiaries.

aerinsol (approved):

PR #400, opened 2020-11-26
Fix Errorhandling in QR pdf generation

HaGuesto:


HaGuesto:
[QR Generation error handling](https://trello.com/c/hdTCI1BH/654-qr-generation-error-handling)

PR #403, opened 2020-12-13
Update foreign keys in db table people_tags

HaGuesto:


HaGuesto:
[People with tags are not possible to delete completely](https://trello.com/c/4PUEQUgE/692-people-with-tags-are-not-possible-to-delete-completely)

PR #405, opened 2021-03-14
add exceptions for qr code generation

HaGuesto:


HaGuesto:
[New generated and printed qr codes are referring to already existing boxes when scanned](https://trello.com/c/pGqSpQk0/709-new-generated-and-printed-qr-codes-are-referring-to-already-existing-boxes-when-scanned)

PR #407, opened 2021-04-24
fix add first base

HaGuesto:


HaGuesto:
[BUG: Demo instance creation is not yet self-service due to error in base creation #small](https://trello.com/c/DxnJeHb5/717-bug-demo-instance-creation-is-not-yet-self-service-due-to-error-in-base-creation-small)

PR #415, opened 2021-05-03
Simplify users screen database queries

jamescrowley:
I’m working on the [Auth0 integration in the old PHP app](https://github.com/boxwise/dropapp/pull/414). Because users can be managed from Auth0, and also from the old PHP app - and ultimately from the new app too - I’m looking at just using Auth0 as the source of truth - otherwise we'll get into messy 2/3-way syncing to keep everything up to date.

The current ‘users’ screen in the PHP app filters which users are visible based on the current users permissions, which makes sense. This should be do-able in Auth0 too - https://auth0.com/docs/extensions/delegated-administration-extension/delegated-administration-hooks/delegated-administration-filter-hook#hook-contract.  - which would then mean both the PHP app and the new python app can also just call the Auth0 getUsers API method too, without having to implement filtering either.

However, the current logic for displaying this view is super complicated, seemingly very redundant, and duplicated across 6 different SQL queries for 3 slightly different screens (active, inactive and deleted users). Therefore, I'm hoping to simplify this.

Given there is no testing here, and I'm not clear on what the intention is in the first place,  I may well be missing something obvious that means I can't do these changes.

I've got a few steps to this, but @HaGuesto if you have any thoughts on the initial simplification - and what, if anything, I'm missing, that would be fab! @hhawilo :wave: just tagging as you might be interested at a high level (more ^^ than the code here itself)

### DIFF
--- a/include/cms_users.php
+++ b/include/cms_users.php
@@ -18,24 +18,14 @@
             addbutton('loginasuser', $translate['cms_users_loginas'], ['icon' => 'fa-users', 'confirm' => true, 'oneitemonly' => true, 'disableif' => true]);
         }
 
-        $camps = db_value(
-            '
-			SELECT GROUP_CONCAT(c.id) 
-			FROM cms_usergroups_camps AS uc, camps AS c 
-			WHERE (NOT c.deleted OR c.deleted IS NULL) AND uc.camp_id = c.id AND uc.cms_usergroups_id = :usergroup',
-            ['usergroup' => $_SESSION['usergroup']['id']]
-        );
-
         // Execution of queries in cms_users_page.php
         $cms_users_lower_level_query = '
 			SELECT u.*, NOT u.is_admin AS visible, g.label AS usergroup, 0 AS preventdelete, 0 as disableifistrue
 			FROM cms_users AS u
 			LEFT OUTER JOIN cms_usergroups AS g ON g.id = u.cms_usergroups_id 
-			LEFT OUTER JOIN cms_usergroups_camps AS uc ON uc.cms_usergroups_id = g.id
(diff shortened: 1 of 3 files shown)


HaGuesto:
@jamescrowley What this deletion leads to is that:
organisations with multiple bases (camps) always show all users of all bases below or equal to your usergroup level (admin, coordinator, volunteer).
Example:
Organisation Boxaid has two Bases (Lesvos and Samos).
Boxaid has a usergroup for coordinators on Samos and coordinators on Lesvos with the right to view the User Mgmt.
A user from the usergroup Samos can then access all the users of Samos and Lesvos even though they should just have access to Samos.

jamescrowley:
thanks @HaGuesto, bit of a brain fail there. Should still be able to simplify this, but I guess would need the user group data in Auth0 for this to work. But if not then we're stuck again trying to regularly sync between Auth0 and the database incase they diverge (particularly if the data is modified in Auth0). Will have more of a think.

aerinsol:
@jamescrowley my gut feeling is that it will be tricky to integrate usergroup permissions into Auth0 because of the way that it closely interacts with the cross-reference tables in the DB - I suppose you could pass it as a parameter from Auth0 and match it but not sure how that would work in terms of loading views.
 
I am also open to re-examining the Boxtribute God permissions if there's a better way to do that.

jamescrowley:
@aerinsol agreed. I'm falling back to just using auth0 for authentication and no authorization. Unfortunately this means we'll have to re-implement logic in the python app for this too, and if anyone manually changes data in Auth0 itself this won't be reflected in the PHP app, but I can't see a way around this for now.

PR #418, opened 2021-05-08
Fix unknown user error message

HaGuesto:


HaGuesto:
[Invalid Email Message doesn't work (and test is not functional) #small #javascript](https://trello.com/c/LjM1n8Az/539-invalid-email-message-doesnt-work-and-test-is-not-functional-small-javascript)

PR #419, opened 2021-05-08
Move domains

HaGuesto:
After checking the code extensively, it turned out to be quite easy.
The important bits are:
- change the gcloud-entry file so that old QR codes from Drop in the Ocean are processed in the right way. (need to be tested)
- change the dispatch.yaml file. (was already deployed to Google cloud --> staging.boxtribute.org already works)
- make sure that the cypress tests run on the right domain.
- update the analytics file.

Next step (once merged):
- add a permanent redirect from staging.boxwise.co to staging.boxtribute.org (@jamescrowley might eed to help me with SSL. I ran into a problem)
- testing, testing, testing


HaGuesto:
[Migrate production app to app.boxtribute.org and demo.boxtribute.org, including all QR codes #large](https://trello.com/c/nHcZ7eCM/563-migrate-production-app-to-appboxtributeorg-and-demoboxtributeorg-including-all-qr-codes-large)

PR #422, opened 2021-07-17
Add a sizegroup with each year for children

HaGuesto:


HaGuesto:
[Create separated children's sizes so that each size is one child year #php #database #small](https://trello.com/c/yUJKC7Jc/729-create-separated-childrens-sizes-so-that-each-size-is-one-child-year-php-database-small)

PR #424, opened 2021-07-20
Second Password Validation and Strength Test

vahidbazzaz:
The following is done:

- [X] Resolved the second password validation
- [X] Test strength of password (it can consist of alphanumerals characters and special character (such as `!@#$%^&*-_=+) and should has minimum of 12 characters lenght includes at least a letter and 2 symbol or numbers)
- [X] Added checkPassword method in tools.php to compare password with second password and checking the strength
- [X] Added custom validation rule in magic.js to validate password strength
- [X] Modified version of magic.js in query string in the cms_footer.php to force browser to update JS cache

### DIFF
--- a/assets/js/magic.js
+++ b/assets/js/magic.js
@@ -429,12 +429,12 @@ $(function() {
         $.validator.addMethod("pwcheck", function(value) {
                 // when a password is not set
                 return (value == "") || ( 
-                    /^[A-Za-z0-9\d=!\-@\^\&\%\#._*]*$/.test(value) // consists of only these
-                    && value.length >= 8 // at least 8 character
-                    && /[a-z]/.test(value) // has a lowercase letter
-                    && /[A-Z]/.test(value) // has a uppercase letter
-                    && /\d/.test(value)) // has a digit
-        }, "Password can consist of alphanumerals characters and special character (such as !@#$%^&*) and should has minimum of 8 characters lenght includes a lower-case letter, an upper-case letter, and a number");
+                    /^[A-Za-z0-9\d`!@#$%^&*\-_=+'\/.,]*$/.test(value) // consists of only these
+                    && value.length >= 12 // at least 8 character
+                    && /[a-zA-Z]/.test(value) // has a letter
+                    && /(?:[^\d`!@#$%^&*\-_=+'\/.,]*[`\d!@#$%^&*\-_=+'\/.,]){2}/.test(value) // has at least 2 symbol or numbers
+                ) 
+        }, "Password can consist of alphanumerals characters and special character (such as `!@#$%^&*-_=+) and should has minimum of 12 characters lenght includes at least a letter and 2 symbol or numbers");
 
         
 
--- a/assets/js/magic.js
+++ b/assets/js/magic.js
@@ -427,16 +427,25 @@ $(function() {
 
 
         $.validator.addMethod("pwcheck", function(value) {
-                // when a password is not set
-                return (value == "") || ( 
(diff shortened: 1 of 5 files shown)


aerinsol:
I'd like to increase the length to 12 and remove the upper/lowercase
requirement, and instead require use of at least 2 symbol or numbers

aerinsol:
See:
https://www.lmgsecurity.com/how-long-should-your-password-be-a-technical-guide-to-a-safe-password-length-policy/

As recent improvements in computing power make 8 character passwords
insufficient.

vahidbazzaz:
@aerinsol I just updated the password strength requirements per your suggestion. Per Auth0 password strength requirement also at least one alphabet  is needed, so I require that as well.

aerinsol (changes requested):
update comment to match 12 char requirement

aerinsol (changes requested):
typo "length"

aerinsol (changes requested):
typo should be "symbols"

aerinsol (changes requested):
Can we separate out the error to the two different use cases, so the user knows which behavior to correct:

1) The password does not match with your previously confirmed password
AND
2) Your password must be at least 12 characters long and include a letter and two symbols or numbers.

aerinsol (changes requested):
request correction to "should be a minimum length of 12 characters, including a letter and two symbols or numbers"

vahidbazzaz:
> Can we separate out the error to the two different use cases, so the user knows which behavior to correct:
> 
> 1. The password does not match with your previously confirmed password
>    AND
> 2. Your password must be at least 12 characters long and include a letter and two symbols or numbers.

Sure, actually its currently work like what you are suggesting; when a user typing any character on the password input, one validator checking if the password is strong enough and likewise when user typing in the confirm password input, it shown error if its not the same as the password.

vahidbazzaz:
> request correction to "should be a minimum length of 12 characters, including a letter and two symbols or numbers"

@aerinsol I think we need to use at least one uppercase and lowercase as Auth0 does not accept common password as well, please check this and then we can finalize the password requirements [https://auth0.com/docs/connections/database/password-strength](url)

aerinsol:
@vahidbazzaz that's the default for the admins dashboard, you mean. I'd prefer to specify the combination of special characters to match:

> Good: at least 8 characters including at least 3 of the following 4 types of characters: a lower-case letter, an upper-case letter, a number, a special character (such as !@#$%^&*).

While keeping it at 12 characters.

aerinsol (changes requested):
We should match the profile password error with the password hints for password creation so we can keep them in sync. 

We can use the same one as done in password creation or simply unify them as a string in a variable, whichever makes the most sense.

vahidbazzaz:
> We should match the profile password error with the password hints for password creation so we can keep them in sync.
> 
> We can use the same one as done in password creation or simply unify them as a string in a variable, whichever makes the most sense.

@aerinsol I think the first suggestion can be done as using the variables in JS can't be done in current setting. Also we you please let me know if the following message is okay to be set for both JS  and backend validation error:

> Your password must be at least 12 characters including at least 3 of the following 4 types of characters: a lowercase letter, an uppercase letter, a number, a special character (such as !@#$%&/=?_.,:;-).

aerinsol:
@vahidbazzaz looks good!

aerinsol (approved):

HaGuesto (approved):
## Code Review

The code looks super clean. Very well done. No complaints at all, @Vahid.
Quite a mastery of regular expressions.

## Functional Review
No problems encountered at all.

PR #434, opened 2021-09-25
Update Auth0 Authentication ADR

vahidbazzaz:
This PR revised the current Auth0 Authentication ADR by adding recent changes and minimal update

### DIFF
--- a/docs/adr/adr_auth0.md
+++ b/docs/adr/adr_auth0.md
@@ -23,16 +23,22 @@ Our handling of user data is far from bullet-proof. We have stumbled upon cases 
 - Google Firebase
 - Auth0
 - Good docs & Supports
-- Comparison Auth0 and Google Firebase 
+- [Comparison Auth0 and Google Firebase](https://docs.google.com/document/d/1sWmS_VDHDx3VDX7JFu78fpU8yMF9eUOHvp2PqJ9_dy0/edit?usp=sharing)
 
 ## Decision
 
-We are going for Auth0 since 
-- we have prior experience with Auth0 in the team, 
-- a first test in an afternoon coding session were satisfying and
-- Auth0 offers a free plan for Open-source projects.
+We are going for Auth0 since
+
+- We have prior experience with Auth0 in the team
+
+- a first test in an afternoon coding session were satisfying
+
+- Auth0 offers a free plan for Open-source projects
+
 - Auth0 is one of the product leaders in CIAM (Visit Gartner for detail at [https://www.g2.com/categories/customer-identity-and-access-management-ciam#grid] - Read recent Leadership Compass Identity API Platforms By KuppingerCole [https://static.carahsoft.com/concrete/files/6315/6840/5267/kuppingercole-leadership-compass-identity-api-platforms.pdf])
+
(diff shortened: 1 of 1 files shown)


aerinsol (changes requested):
Two of the key points in your doc comparison is that pricing is an issue once we hit 10K (users? verifications? what's the rate limit unit here?) 

So we should capture that point under "harder".

vahidbazzaz:
> Two of the key points in your doc comparison is that pricing is an issue once we hit 10K (users? verifications? what's the rate limit unit here?)
> 
> So we should capture that point under "harder".

@aerinsol Sure, thanks that make sense and I will update it accordingly.

docs/adr/adr_auth0.md
```diff
@@ -35,13 +35,21 @@ We are going for Auth0 since
 - Auth0 has good documentation and active community support (https://auth0.com/docs - https://community.auth0.com/)
 - Auth0 and Google Firebase comparison in detail [https://docs.google.com/document/d/1sWmS_VDHDx3VDX7JFu78fpU8yMF9eUOHvp2PqJ9_dy0/edit?usp=sharing]
```
aerinsol, line 36: @vahidbazzaz as mentioned before please put this in "Considered Options"

docs/adr/adr_auth0.md
```diff
@@ -35,13 +35,21 @@ We are going for Auth0 since
...
+### Easier:
+- We can almost drop a whole user flow
+
+### More difficult:
+
+- Switching between mobile and desktop application
+
+- Reaching the total users (10K) or API rate limits, and how they impact pricing 
+
+- Changes to the current Auth0 enterprise agreement for open source project
```
aerinsol: What are the changes and why should we be concerned?

docs/adr/adr_auth0.md
```diff
@@ -35,13 +35,21 @@ We are going for Auth0 since
...
-###Easier:
-We can almost drop a whole user flow.
+### Easier:
+- We can almost drop a whole user flow
+
+### More difficult:
+
+- Switching between mobile and desktop application
+
+- Reaching the total users (10K) or API rate limits, and how they impact pricing
```
aerinsol: Should this say "becomes much more expensive at a rate of $XXX instead of $YYY from comparable solutions". The goals of this ADR Is that people can read this and understand the conclusion and how bad the problem is, without digging around.

aerinsol (changes requested):
Still not totally clear.

docs/adr/adr_auth0.md
vahidbazzaz: @aerinsol Please recheck the most recent version since you seem to have commented on an outdated version of the file. I already addressed that concern regarding pricing.

docs/adr/adr_auth0.md
vahidbazzaz: @aerinsol Could you please confirm that the following is aligned with what you mentioned:

- Good Docs and Supports
- Comparison Auth0 and Google Firebase

aerinsol (approved):

PR #435, opened 2021-09-27
Remove recurring login prompt

HaGuesto:


HaGuesto:
[User has to login for each box they register](https://trello.com/c/4qGvNL1w/755-user-has-to-login-for-each-box-they-register)

PR #438, opened 2021-10-29
Upgrade PHP version to 7.4.25

vahidbazzaz:
This PR is for [upgrading PHP to 7.4.25](https://trello.com/c/Nafue8Kp), which is required by some libraries.

- The composer.json file has been updated
- No issues were found while running the Cypress tests

### DIFF
--- a/include/people.php
+++ b/include/people.php
@@ -426,6 +426,7 @@ Tracer::inSpan(
                 }
 
                 break;
+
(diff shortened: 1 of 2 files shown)


HaGuesto (approved):
@vahidbazzaz thanks! I updated the app.yaml for Google App Engine and the version of php in CircleCI.

PR #440, opened 2021-11-06
Quick fix for the boxes and beneficiaries loading issue

vahidbazzaz:
This PR limits the number of records on the boxes and beneficiaries list in order to resolve the performance issue 

- A new setting added to the list.php to set `maxlimit`
- A new block for notification added to the footer
- Both boxes and beneficiaries list limited to displaying 500 records
- Notification added to the  boxes and beneficiaries page


HaGuesto (approved):
Nice solution for using noty in the footer @vahidbazzaz. 

I changed the PR that the notification is only shown if the LIMIT actually limits the list to 500. 

Please do not pus the cache of the formatter next time. That was a bit annoying to review.

PR #441, opened 2021-11-08
Enable zero count for items on the mobile ui

vahidbazzaz:
This PR is minor change in javascript validation to [enable zero count for items](https://trello.com/c/Y5kV3VTg) on the mobile UI.


HaGuesto (approved):
works. Thanks @vahidbazzaz !

PR #442, opened 2021-11-08
Box number generation refactored

vahidbazzaz:
- Increase the box id length to 8 digits
- Prevent the first digit from being zero
- Limit the box id lookup to 10 times


HaGuesto (approved):
Looks good @vahidbazzaz!

PR #444, opened 2021-11-18
Fixes reactivating dependent without family head

vahidbazzaz:
This PR fixes a reported issue, which let the user reactivate the dependent without the active family head.

The changes are as follows:

- Update people_deactivated.php to verify that the selected beneficiaries have a family head; if they do not, they are prevented from being reactivated
- Run the Cypress test for beneficiary management and all the test cases passed


HaGuesto (approved):
well done @vahidbazzaz !

I checked and the undelete for the family_head also recursively undeletes any child. I deactivated that.

Also I made the page reload after reactivation to update the icons.

PR #445, opened 2021-11-24
Bulk tag removal added to the manage beneficiaries

vahidbazzaz:
The PR includes a feature that allows users to remove beneficiary tags in bulk.

Here is what is done:

- Added new button in the list bar to remove selected tags from checked beneficiaries 
- Set a fixed limit of 500 records for bulk removing to prevent performance and misuse 
- All the Cypress test for the beneficiaries management has run and passed


HaGuesto (approved):
@vahidbazzaz 
I cleaned it up a bit. Looks good. tested it thoroughly and checked the code.

PR #448, opened 2021-11-29
Resolved issue with total boxes

vahidbazzaz:
This is a hotfix for displaying incorrect total numbers of boxes and items in the manage box


vahidbazzaz:
We agreed to remove the limit from the manage box.

PR #450, opened 2021-11-30
Quick fix for sorting columns in Manage Boxes

vahidbazzaz:
This PR contains the quick fix for the issue of sorting in manage box:

- Added a button to show all
- When the show all button is clicked, sorting is enabled


vahidbazzaz:
Agreed: remove limit from manage box

PR #452, opened 2021-12-04
Enabling the sending of zero tokens

vahidbazzaz:
This PR fix the issue that prevent sending zero tokens to all.

### DIFF
--- a/include/give2all.php
+++ b/include/give2all.php
@@ -12,7 +12,7 @@
                 }
         */
         // validate tokens input value
-        if (!preg_match('/[1-9]\d*/', $_POST['dropsadult']) || !preg_match('/[1-9]\d*/', $_POST['dropschild'])) {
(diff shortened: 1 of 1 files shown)


aerinsol:
@vahidbazzaz thank you for the hotfix. I've merged this into master, but I would like you to add a code comment here that this allowance for 0 is specifically for the "reset tokens" feature, as requested by Darbazar. This way in the future when we need to edit this, it will be more maintainable.

Can we add this?

vahidbazzaz:
> @vahidbazzaz thank you for the hotfix. I've merged this into master, but I would like you to add a code comment here that this allowance for 0 is specifically for the "reset tokens" feature, as requested by Darbazar. This way in the future when we need to edit this, it will be more maintainable.
> 
> Can we add this?
 @aerinsol Sure, but I think a new PR is required since the current one is already merged and closed.

aerinsol:
https://trello.com/c/Um44yV7y

Sounds good @vahidbazzaz. Just add a new one and link it to the Trello card above.

PR #454, opened 2021-12-20
Update Auth0 PHP SDK 8

vahidbazzaz:
This PR contains changes that are needed to [integrate the new Auth0 PHP SDK 8.0.5](https://trello.com/c/9JJTFeDn). This library requires PHP version higher than 7.4, so this PR includes an update to PHP version as well.

- The last login date is now updated at login time with the value of the user token IAT (issued at) and stored in the local database
- Session.php has undergone major changes, and some cleanup has been done as part of the new SDK integration
- The Cypress background login has been changed to work with the new SDK
- Except for test 1.1, 1.2, and 2.9.9 (last part) which require Auth0 login redirect, all other tests are working as expected
- I've disabled the tests that required the Auth0 login UI, as the cross origin error and the new changes in Auth0 SDK even with the form_post response type no longer work in the development environment with an insecure protocol (http). All other tests still work as expected.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,6 +68,7 @@ jobs:
             \$settings['auth0_domain'] = '${<< parameters.envVariablePrefix >>AUTH0_DOMAIN}';
             \$settings['auth0_client_id'] = '${<< parameters.envVariablePrefix >>AUTH0_CLIENT_ID}';
             \$settings['auth0_client_secret'] = '${<< parameters.envVariablePrefix >>AUTH0_CLIENT_SECRET}';
+            \$settings['auth0_cookie_secret'] = '${<< parameters.envVariablePrefix >>AUTH0_COOKIE_SECRET}';
             \$settings['auth0_redirect_uri'] = '${<< parameters.envVariablePrefix >>AUTH0_REDIRECT_URI}';" > library/config.php
           echo "paths:
             migrations: '%%PHINX_CONFIG_DIR%%/db/migrations'
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,6 +67,7 @@ jobs:
             \$settings['test_pwd'] = '${CYPRESS_TEST_PWD}';
             \$settings['auth0_domain'] = '${<< parameters.envVariablePrefix >>AUTH0_DOMAIN}';
             \$settings['auth0_client_id'] = '${<< parameters.envVariablePrefix >>AUTH0_CLIENT_ID}';
+            \$settings['auth0_db_connection_id'] = '${<< parameters.envVariablePrefix >>AUTH0_DB_CONNECTION_ID}';
             \$settings['auth0_client_secret'] = '${<< parameters.envVariablePrefix >>AUTH0_CLIENT_SECRET}';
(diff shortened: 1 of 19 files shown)


aerinsol (approved):

library/lib/session.php
```diff
@@ -149,18 +215,21 @@ function updateAuth0UserFromDb($user_id, $set_pwd = false)
...
+    // user doesn't exist, so try creating it instead
+    if (404 === $response->getStatusCode()) {
+        $auth0UserData['user_id'] = preg_replace('/auth0\|/', '', $auth0UserId);
+        $auth0UserData['password'] = generateSecureRandomString(); // user will need to reset password anyway
+        $response = $mgmtAPI->users()->create($settings['auth0_db_connection_id'], $auth0UserData);
+        // the status code will be 201 if the user created successfully
+        if (201 !== $response->getStatusCode()) {
+            throw new Exception($response->getStatusCode(), $response->getReasonPhrase());
         }
+    } elseif (200 !== $response->getStatusCode()) {
```
HaGuesto, line 229: what case is this @vahidbazzaz ?

library/lib/session.php
```diff
@@ -241,18 +310,39 @@ function logout()
...
+    exit;
+}
+
+function loginWithRedirect()
+{
+    global $settings;
+    $auth0 = getAuth0($settings);
+
+    // Clear the local session.
+    $auth0->clear();
```
HaGuesto, line 340: u think this will resolve the Invalid State errors we always got from Auth0? @vahidbazzaz

library/lib/session.php
```diff
@@ -396,12 +484,14 @@ function getAuth0User($userId)
     try {
         global $settings;
         $mgmtAPI = getAuth0Management($settings);
-        $user = $mgmtAPI->users()->get('auth0|'.intval($userId));
+        $response = $mgmtAPI->users()->get('auth0|'.intval($userId));
 
-        return $user;
-    } catch (GuzzleHttp\Exception\ClientException $e) {
+        if (HttpResponse::wasSuccessful($response)) {
```
HaGuesto, line 489: I don't get what you are doing here @vahidbazzaz

.circleci/config.yml
```diff
@@ -67,7 +67,9 @@ jobs:
             \$settings['test_pwd'] = '${CYPRESS_TEST_PWD}';
             \$settings['auth0_domain'] = '${<< parameters.envVariablePrefix >>AUTH0_DOMAIN}';
             \$settings['auth0_client_id'] = '${<< parameters.envVariablePrefix >>AUTH0_CLIENT_ID}';
+            \$settings['auth0_db_connection_id'] = '${<< parameters.envVariablePrefix >>AUTH0_DB_CONNECTION_ID}';
```
HaGuesto, line 70: why do we need this parameter? @vahidbazzaz I do not see where it is used.

HaGuesto (commented):
u need to write in README how to set the cookie secret
image

.circleci/config.yml
HaGuesto: needed to update user

PR #461, opened 2022-02-11
Configure dispatch URL for api-staging service

pylipp:
Recap for my understanding:
1. When this PR is merged, the URL `api-staging.boxtribute.org` will be online. But calling it will probably give an empty page (or a 404?)
1. In the `boxtribute` repo, I will create a dedicated `app-api-staging.yaml` file to configure the new service `api-staging` with a single handler
1. Also I'll have to extend `.circleci/config.yml` there to actually deploy to that service

Is that correct?

### DIFF
--- a/dispatch.yaml
+++ b/dispatch.yaml
@@ -26,6 +26,8 @@ dispatch:
   service: v2-staging
 - url: "staging.boxtribute.org/v2*"
   service: v2-staging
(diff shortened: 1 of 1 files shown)


vahidbazzaz (approved):

vahidbazzaz:
@pylipp I think you need to add `/*` at the end of the url so it should be `api-staging.boxtribute.org/*`

PR #463, opened 2022-02-16
Add dispatchers for api-demo and -production services

pylipp:
In combination with boxwise/boxtribute#170. All services should be deployed there first, then the dispatching should be activated.


HaGuesto (approved):
looks fine frome my side. u can merge if u want

PR #464, opened 2022-02-23
Integration of AuthZ action permissions in Dropapp 

vahidbazzaz:
This PR contains major changes to integrate Auth0 authorization in Dropapp also enabling custom login url:
- creation of standard user groups in dropapp and also addition of roles in auth0 using cross table cms_usergroups_roles
- syncing users roles with auth0
- updating of action permission in auth0 
- disabling editing of standard user groups
- disabling deletion of bases and organization


vahidbazzaz:
[Implement roles/permissions/db permissions in dropapp](https://trello.com/c/HgKoMDov/811-implement-roles-permissions-db-permissions-in-dropapp)

vahidbazzaz:
NOTE: We need to make sure the test users aren't affected by this PR (synchronizing existing roles with auth0)

aerinsol:
Specifically Hermine test users for the production read access API

db/seeds/ActionPermissions.php
```diff
@@ -0,0 +1,40 @@
...
+    public function getDependencies()
+    {
+        return [
+            'Minimal',
+        ];
+    }
+
+    public function run()
+    {
+        $this->execute('UPDATE cms_functions SET action_permissions = "checkout_beneficiaries" WHERE id=87');
```
HaGuesto, line 16: note for Hans Review: do we need this in the migration?

include/camps_edit.php
```diff
@@ -48,15 +58,15 @@
     addfield('line', '', '', ['tab' => 'general']);
 
     addfield('title', 'Features', '', ['tab' => 'general']);
-    addfield('select', 'Functions available for this base', 'functions', ['width' => 6, 'tab' => 'general', 'multiple' => true, 'query' => '
```
HaGuesto, line 51: note for Hans review: How does the disabling work?

include/cms_usergroups.php
```diff
@@ -62,7 +62,10 @@
 
             case 'delete':
```
HaGuesto, line 63: Hans review: check all cases for usergroups edit

reseed-auth0-roles.php
```diff
@@ -0,0 +1,87 @@
+<?php
```
HaGuesto, line 1: put this file into reseed-auth0

library/lib/auth0.php
```diff
@@ -1 +1,848 @@
...
+        }
+    } catch (Auth0Exception $e) {
+        // user doesn't exist in auth0
+        if (404 == $e->getCode()) {
+            return null;
+        }
+    }
+}
+/**
+ * Getting all avialble End-Points in Auth0.
```
HaGuesto, line 587: remove resource stuff if it is not needed.

library/constants.php
```diff
@@ -0,0 +1,536 @@
+<?php
+
+// static - temp
+
+$menusToActions = [
```
HaGuesto, line 5: @vahidbazzaz is this constant still needed?

library/lib/auth0.php
```diff
@@ -1 +1,791 @@
...
+        $mgmtAPI = getAuth0Management($settings);
+        $response = $mgmtAPI->roles()->create($roleName);
+
+        if (HttpResponse::wasSuccessful($response)) {
+            $res = HttpResponse::decodeContent($response);
+
+            return $res;
+        }
+    } catch (Auth0Exception $e) {
+        // user doesn't exist in auth0
```
HaGuesto, line 419: I do not like the error handling here. Looks very copy/paste

library/lib/auth0.php
```diff
@@ -1 +1,791 @@
...
+                ugr.cms_usergroups_id=:userGroupsId', ['userGroupsId' => $dbUserData['cms_usergroups_id']]);
+
+    usleep(1000);
+    // assign user roles into auth0
+    $roles = [];
+    foreach ($dbUserRoles as $role) {
+        $roles[] = $role['auth0_role_id'];
+    }
+    if (sizeof($roles) > 0) {
+        assignRolesToUser($auth0UserId, $roles);
```
HaGuesto, line 173: It feels like it always adds the roles before checking if they are already assigned.

HaGuesto (changes requested):
# Review
@vahidbazzaz I decided to check the different bullet points one by one for the functionalreview. The code review I comment inline

## functional review:
(0.) db migrations and seeds
migration works and can be reverted. It looks good.
For the seed I had to add some permissions, but it looks good now.

1. Remove access of users to usergroups menu to not get the mapping to roles out of sync
That's done in production. I just checked again.

2. create all roles in Auth0 based on the dev seed db (bases, organisation). This should be about 7 roles per base. 
In dev all roles are there. In staging there are some missing, but we can test the script you wrote once it is merged.
At some point you need to explain to me again why there is only one administrator role for all orgs.

3. create functions for dynamic creation of roles in Auth0 if an organisation is created. (Admin role only) 
write function when first base is created. There should be 7 usergroups in dropapp created and 5 roles in Auth0. DO NOT forget to write entries for usergroups in cms_usergroups_functions and cms_usergroups_camps

This works fine. I'm a bit concerned if we should really use the global variable $menusToActions or just query the menues from action_permissions in cms_functions.
We also need to go through the error handling in some of the functions in auth0.php

4. create functions for dynamic creation of roles in Auth0 if an base is created.
write function when second to nth base is created. There should be 6 usergroups in dropapp created (no admin) and 5 roles in Auth0. DO NOT forget to write entries for usergroups in cms_usergroups_functions and cms_usergroups_camps and set userlevel. DO NOT forget to assign a new base to the admin usergroup
Works fine. Same comments as for 3.)

5. create db table cms_usergroups_roles which tracks the mapping of usergroups to Auth0 roles.
works fine.

6. add roles to cms_functions (write db migration for production and add a form field in cms_functions_edit to access this field)
This works and I added also a column to show the permissions in the table of cms_functions.

8. create a test function which can be run by a cron job that checks every night if all roles exist in Auth0 that are needed according to the db bases and organisation and cms_usergroups_roles. (check if roles are in sync to bases and organisations).
This is missing.

9. when a user is created, edited or deleted assign the roles according to cms_usergroups_roles in Auth0.
This only adds new roles to a user in Auth0, but does not remove old ones.

11. update cron job test if all users are in sync between Auth0 and the db to now also check if if the users roles are in sync with the usergroup of the dropapp.
This works, but we need to add a seed for cms_usergroups_roles.

12. disable renaming of usergroups
This does not work.

13. disable delete of usergroups
This works

14. if base is renamed, rename the user group names
This works. The problem is that if old usergroup names are renamed it creates problems.

15. disable delete of bases
I edited this a bit.

16. disable delete of organisations 
I edited this a bit.

17. create all standard auth0 roles, which are created usually new bases/organisations, also for each existing base/organisation; we do this by running parts of the auth0-reseed cron job (not the password reset) against production; for this we create a script similiar to auth0-reseed, add it as an entrypoint in gcloud-entry and add a line in cron.yaml. when it is live we can trigger the script through google appengine.
I have not checked this.

db/auth0/create-auth0-roles-usergroups.php
```diff
@@ -0,0 +1,49 @@
...
+    foreach (array_values($rolesToActions[$k]) as $k) {
+        $methods[] = $k;
+    }
+}
+
+$methods = array_unique($methods);
+
+// add action permissions in the auth0 API end-point
+// boxtribute-dev-api --> id: 5ef3760527b0da00215e6209
+updateResources('5ef3760527b0da00215e6209', $methods);
```
HaGuesto, line 32: this needs to be replaced by a $settings variable which need to be set in CircleCI

db/auth0/create-auth0-roles-usergroups.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+if (!array_key_exists('HTTP_X_APPENGINE_CRON', $_SERVER) && 'dropapp_dev' != $settings['db_database']) {
+    throw new Exception('Not called from AppEngine cron service');
+}
+    $permittedDatabases = ['dropapp_dev', 'dropapp_demo', 'dropapp_staging'];
```
HaGuesto, line 6: remove these line

HaGuesto:
Summary: What is left to be done?

1. If a user is edited and a different usergroup is assigned, the role in auth0 schould not just be assigned, but also the old roles removed.
2. disable name field in usergroups_edit
3. fix renaming bases of already existing bases.
4. fix [db/auth0/create-auth0-roles-usergroups.php](https://github.com/boxwise/dropapp/pull/464/files#diff-f4f607b96340f67e911c92592787a783aa15c35949d5569bf5468b8d795fc816) by adding cron job and entrypoint
5. @HaGuesto decide how when to run the scripts reseed-auth0 and resee-roles-auth0.


What do we move to a different ticket?
- We need to rework the dev seed so that the usergroups in the seed are connected with the roles in Auth0 on local dev environments and staging.
- maybe exchange $menusToActions by a query to cms_functions.

db/auth0/create-auth0-roles-usergroups.php
vahidbazzaz: @HaGuesto I remove database name checking but keep the first few lines to ensure that this script only run via cron job.

db/auth0/create-auth0-roles-usergroups.php
vahidbazzaz: @HaGuesto I have added environment variables in the CircleCi and now using $settings to bind the value

library/lib/auth0.php
vahidbazzaz: @HaGuesto  I fixed this by removing the roles at first hands and then adding new roles.

vahidbazzaz:
@HaGuesto Thanks for the review, I made the changes as per your suggestions.

library/lib/auth0.php
vahidbazzaz: @HaGuesto I have improved the exception handling

library/constants.php
vahidbazzaz: @HaGuesto I will implement query version in  the different ticket as we discussed

HaGuesto (approved):
Hey @vahidbazzaz Vahid, locally all tests work and I do not see any big flaws on first sight. I'm gonna merge on master and will do some more tests there.

/Hans

PR #469, opened 2022-03-15
Prevent Duplicates QR labels Print

vahidbazzaz:
This PR adds a page number to the footer on the PDF generation to make sure it is not a printer repeat error.


vahidbazzaz:
[There are duplicate QR codes generated if you generate more than 20 or 30 at once](https://trello.com/c/OjWZzsGA/850-there-are-duplicate-qr-codes-generated-if-you-generate-more-than-20-or-30-at-once)

pdf/pdf.php
```diff
@@ -12,6 +12,20 @@ public function Print($text, $x = 0)
         $this->Text($this->X + $x, $this->Y, $text);
     }
 
+    // related to this trello: https://trello.com/c/OjWZzsGA
+    public function Footer()
+    {
+        // Go to 1.5 cm from bottom
+        $this->SetY(-10);
+        // Select Arial italic 8
+        $this->SetFont('Arial', 'I', 6);
```
aerinsol, line 21: Open Sans (per our brand identity guidelines) please.

aerinsol (changes requested):

pdf/pdf.php
vahidbazzaz: @aerinsol Since we were previously using Helvetica in the PDF, I have incorporated Open Sans fonts in accordance with your recommendation and our brand identity guidelines, and updated the file accordingly.

aerinsol (approved):

aerinsol (approved):

PR #477, opened 2022-04-07
Fix for QR Label Footer Text Alignment

vahidbazzaz:
The PR fixes a minor issue with the display of footer text in QR labels.


vahidbazzaz:
[There are duplicate QR codes generated if you generate more than 20 or 30 at once #release1.16](https://trello.com/c/OjWZzsGA)

aerinsol (approved):

PR #478, opened 2022-04-14
Bulk Delete for Tags Removal in Manage Beneficiaries

vahidbazzaz:
This PR contains minor fixes to optimize removing tags query using bulk delete.


vahidbazzaz:
[Further investigation and Refactoring of Dropapp Performance Issues](https://trello.com/c/g24mIVb8)

aerinsol (approved):
FYI: Normally I wouldn't agree with hardcoding table names and primary keys here, but given this is a performance workaround on a legacy code base, it's fine.

PR #488, opened 2022-06-05
Create composite index for column tablename and record_id in history table

vahidbazzaz:
This PR contains db clean up for history table; two composite index introduced based on the analysis of the Drop App queries.


vahidbazzaz:
[Create composite index for column tablename and record_id in history table](https://trello.com/c/TWjxrQ7s)

HaGuesto (approved):

PR #491, opened 2022-06-17
Fix product DB migrations

HaGuesto:


HaGuesto:
[Database cleanup Product MYSQL table](https://trello.com/c/G1GwpFXC/898-database-cleanup-product-mysql-table)

PR #497, opened 2022-07-01
Create Shipment and Transfers Tables

vahidbazzaz:
This PR contains the migrations required for creation of shipment and transfer tables. 

Related PR: https://github.com/boxwise/boxtribute/pull/285

### DIFF
--- a/db/migrations/20220701094045_add_transfer_agreement_table.php
+++ b/db/migrations/20220701094045_add_transfer_agreement_table.php
@@ -1,0 +1,104 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class AddTransferAgreementTable extends AbstractMigration
+{
+    public function change()
+    {
+        $transferAgreement = $this->table('transfer_agreement', [
+            'id' => true,
+            'primary_key' => ['id'],
+            'engine' => 'InnoDB',
+            'signed' => false,
+            'encoding' => 'utf8mb4',
+            'collation' => 'utf8_general_ci',
+            'comment' => '',
+            'row_format' => 'DYNAMIC',
+        ]);
+
+        $transferAgreement->addColumn('source_organisation_id', 'integer', [
+            'null' => false,
+            'signed' => false,
+            'after' => 'id',
+        ])
+            ->addColumn('target_organisation_id', 'integer', [
+                'null' => false,
+                'signed' => false,
+                'after' => 'source_base_id',
+            ])
+            ->addColumn('status', 'string', [
+                'null' => false,
+                'default' => 'UnderReview',
+                'limit' => 255,
+                'after' => 'target_organisation_id',
+            ])
+            ->addColumn('type', 'string', [
+                'null' => false,
+                'limit' => 255,
+                'after' => 'status',
+            ])
+            ->addColumn('requested_on', 'datetime', [
+                'null' => false,
+                'after' => 'type',
+            ])
+            ->addColumn('requested_by', 'integer', [
+                'null' => false,
+                'signed' => false,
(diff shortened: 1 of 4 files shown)


vahidbazzaz:
[BUG/ISSUE: Shipment and transfers tables do not currently exist in the live staging / production databases](https://trello.com/c/YB02E4tM)

db/migrations/20220701094045_add_transfer_agreement_table.php
```diff
@@ -0,0 +1,104 @@
...
+                'after' => 'transfer_agreement_id',
+            ])
+            ->addForeignKey('source_organisation_id', 'organisations', 'id', [
+                'update' => 'CASCADE',
+            ])
+            ->addForeignKey('target_organisation_id', 'organisations', 'id', [
+                'update' => 'CASCADE',
+            ])
+            ->addForeignKey('requested_by', 'cms_users', 'id', [
+                'update' => 'CASCADE',
```
pylipp, line 87: Does it make sense to not have the `delete` option set? I know you followed the peewee model definition so I wonder what I was thinking to not define `delete` there :thinking:

db/migrations/20220701094148_add_shipment_table.php
```diff
@@ -0,0 +1,107 @@
...
+                'null' => false,
+                'signed' => false,
+                'after' => 'source_base_id',
+            ])
+            ->addColumn('transfer_agreement_id', 'integer', [
+                'null' => false,
+                'signed' => false,
+                'after' => 'target_base_id',
+            ])
+            ->addColumn('status', 'string', [
```
pylipp, line 34: See above. Requesting to change to `state` here and in line 42 :)

db/migrations/20220701094045_add_transfer_agreement_table.php
pylipp: Ahh never mind, the `requested_by` field is defined as `NOT NULL`, so `ON DELETE SET NULL` does not make sense.

db/migrations/20220701094045_add_transfer_agreement_table.php
```diff
@@ -0,0 +1,104 @@
...
+            'null' => false,
+            'signed' => false,
+            'after' => 'id',
+        ])
+            ->addColumn('target_organisation_id', 'integer', [
+                'null' => false,
+                'signed' => false,
+                'after' => 'source_base_id',
+            ])
+            ->addColumn('status', 'string', [
```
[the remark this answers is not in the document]
vahidbazzaz: @pylipp if we want to enforce ENUM values, then we need to use it as DB constraint and we can use Check constraint but is not supported in this version Phinx and then we need to run custom query for it. So I think it make sense to leave it as string with default value.

db/migrations/20220701094045_add_transfer_agreement_table.php
```diff
@@ -0,0 +1,104 @@
...
+            ->addColumn('valid_from', 'datetime', [
+                'null' => false,
+                'after' => 'terminated_by',
+            ])
+            ->addColumn('valid_until', 'datetime', [
+                'null' => true,
+                'after' => 'valid_from',
+            ])
+            ->addColumn('comment', 'text', [
+                'null' => false,
```
pylipp, line 77: I wonder whether comment should be nullable, just as in the `stock` table. What do you think?

vahidbazzaz: @pylipp thanks, thats totally make sense, I will make it nullable, please update this in your end too  ;)

pylipp (approved):

PR #499, opened 2022-07-23
Part 1 - Mobile Distro DB Migrations

vahidbazzaz:
This PR contains the migrations needed for round 1 of the mobile distro feature. 

Ref. doc: https://docs.google.com/document/d/144pHmdgkusbnbSgatyHAeeU34gFiWLYOI2puT_CA6ws/edit#

### DIFF
--- a/db/migrations/20220722085453_dirstro_events_table.php
+++ b/db/migrations/20220722085453_dirstro_events_table.php
@@ -1,0 +1,83 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class DirstroEventsTable extends AbstractMigration
+{
+    public function change()
+    {
+        $distroEvents = $this->table('distro_events', [
+            'id' => true,
+            'primary_key' => ['id'],
+            'engine' => 'InnoDB',
+            'signed' => false,
+            'encoding' => 'utf8mb4',
+            'collation' => 'utf8_general_ci',
+            'comment' => '',
+            'row_format' => 'DYNAMIC',
+        ]);
+        $distroEvents->addColumn('name', 'string', [
+            'null' => true,
+            'default' => null,
+            'limit' => '255',
+        ])
+            ->addColumn('planned_start_date_time', 'datetime', [
+                'null' => false,
+                'after' => 'name',
+            ])
+            ->addColumn('planned_end_date_time', 'datetime', [
+                'null' => false,
+                'after' => 'planned_start_date_time',
+            ])
+            ->addColumn('location_id', 'integer', [
+                'null' => false,
+                'signed' => false,
+                'after' => 'planned_start_date_time',
+            ])
+            ->addColumn('state', 'string', [
+                'null' => false,
+                'default' => 'Planning',
+                'limit' => '255',
+                'after' => 'location_id',
+            ])
+            ->addColumn('created_on', 'datetime', [
+                'null' => false,
+                'after' => 'state',
+            ])
+            ->addColumn('created_by', 'integer', [
+                'null' => false,
+                'signed' => false,
+                'after' => 'created_on',
+            ])
+            ->addColumn('deleted_on', 'datetime', [
+                'null' => true,
+                'after' => 'created_by',
+            ])
+            ->addColumn('last_modified_on', 'datetime', [
+                'null' => true,
+                'after' => 'deleted_on',
+            ])
+            ->addColumn('last_modified_by', 'integer', [
+                'null' => true,
+                'signed' => false,
+                'after' => 'last_modified_on',
+            ])
+            ->addForeignKey('location_id', 'locations', 'id', [
+                'delete' => 'RESTRICT', 'update' => 'CASCADE',
+            ])
+            ->addForeignKey('created_by', 'cms_users', 'id', [
+                'update' => 'CASCADE',
+            ])
+            ->addForeignKey('last_modified_by', 'cms_users', 'id', [
+                'delete' => 'SET_NULL', 'update' => 'CASCADE',
+            ])
+            ->addIndex(['planned_start_date_time'], ['name' => 'distro_events_planned_start_date_time'])
+            ->addIndex(['planned_end_date_time'], ['name' => 'distro_events_planned_end_date_time'])
+            ->addIndex(['state'], ['name' => 'distro_events_state'])
+            ->addIndex(['location_id'], ['name' => 'distro_events_location_id'])
+            ->addIndex(['created_by'], ['name' => 'distro_events_created_by'])
+            ->addIndex(['last_modified_by'], ['name' => 'distro_events_last_modified_by'])
+            ->create()
+        ;
+    }
+}
--- a/db/migrations/20220722085453_dirstro_events_table.php
+++ b/db/migrations/20220722085453_dirstro_events_table.php
@@ -9,10 +9,7 @@ class DirstroEventsTable extends AbstractMigration
         $distroEvents = $this->table('distro_events', [
             'id' => true,
             'primary_key' => ['id'],
-            'engine' => 'InnoDB',
             'signed' => false,
-            'encoding' => 'utf8mb4',
-            'collation' => 'utf8_general_ci',
             'comment' => '',
             'row_format' => 'DYNAMIC',
         ]);
@@ -74,9 +71,6 @@ class DirstroEventsTable extends AbstractMigration
             ->addIndex(['planned_start_date_time'], ['name' => 'distro_events_planned_start_date_time'])
             ->addIndex(['planned_end_date_time'], ['name' => 'distro_events_planned_end_date_time'])
             ->addIndex(['state'], ['name' => 'distro_events_state'])
-            ->addIndex(['location_id'], ['name' => 'distro_events_location_id'])
-            ->addIndex(['created_by'], ['name' => 'distro_events_created_by'])
-            ->addIndex(['last_modified_by'], ['name' => 'distro_events_last_modified_by'])
             ->create()
         ;
     }
--- a/db/migrations/20220722085453_dirstro_events_table.php
+++ b/db/migrations/20220722085453_dirstro_events_table.php
@@ -9,7 +9,10 @@ class DirstroEventsTable extends AbstractMigration
         $distroEvents = $this->table('distro_events', [
             'id' => true,
             'primary_key' => ['id'],
+            'engine' => 'InnoDB',
             'signed' => false,
+            'encoding' => 'utf8',
+            'collation' => 'utf8_general_ci',
             'comment' => '',
             'row_format' => 'DYNAMIC',
         ]);
--- a/db/migrations/20220722085453_dirstro_events_table.php
+++ b/db/migrations/20220722085453_dirstro_events_table.php
@@ -42,7 +42,7 @@ class DirstroEventsTable extends AbstractMigration
                 'after' => 'state',
             ])
             ->addColumn('created_by', 'integer', [
-                'null' => false,
+                'null' => true,
                 'signed' => false,
                 'after' => 'created_on',
             ])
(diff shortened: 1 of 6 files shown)


vahidbazzaz:
[Mobile Distro: Needed DB Migrations - Round 1](https://trello.com/c/nEidtRbT)

db/migrations/20220722085453_dirstro_events_table.php
```diff
@@ -0,0 +1,83 @@
...
+use Phinx\Migration\AbstractMigration;
+
+class DirstroEventsTable extends AbstractMigration
+{
+    public function change()
+    {
+        $distroEvents = $this->table('distro_events', [
+            'id' => true,
+            'primary_key' => ['id'],
+            'engine' => 'InnoDB',
```
HaGuesto, line 12: Why did u define engine, encoding and collation? The defaults should be fine. And encoding should be utf8, right?

db/migrations/20220722085453_dirstro_events_table.php
```diff
@@ -0,0 +1,83 @@
...
+            'comment' => '',
+            'row_format' => 'DYNAMIC',
+        ]);
+        $distroEvents->addColumn('name', 'string', [
+            'null' => true,
+            'default' => null,
+            'limit' => '255',
+        ])
+            ->addColumn('planned_start_date_time', 'datetime', [
+                'null' => false,
```
HaGuesto, line 25: If Null is falsche, what is the default?

db/migrations/20220722085453_dirstro_events_table.php
```diff
@@ -0,0 +1,83 @@
...
+                'after' => 'name',
+            ])
+            ->addColumn('planned_end_date_time', 'datetime', [
+                'null' => false,
```
HaGuesto, line 29: If Null is false, what is the default?

db/migrations/20220722085453_dirstro_events_table.php
```diff
@@ -0,0 +1,83 @@
...
+                'after' => 'planned_start_date_time',
+            ])
+            ->addColumn('state', 'string', [
+                'null' => false,
+                'default' => 'Planning',
+                'limit' => '255',
+                'after' => 'location_id',
+            ])
+            ->addColumn('created_on', 'datetime', [
+                'null' => false,
```
HaGuesto, line 44: If Null is falsche, what is the default?

db/migrations/20220723052618_dirstro_events_packing_list_entries.php
```diff
@@ -0,0 +1,85 @@
...
+    public function change()
+    {
+        $distroEventsPackingListEntries = $this->table('distro_events_packing_list_entries', [
+            'id' => true,
+            'primary_key' => ['id'],
+            'engine' => 'InnoDB',
+            'signed' => false,
+            'encoding' => 'utf8mb4',
+            'collation' => 'utf8_general_ci',
+            'comment' => '',
```
HaGuesto, line 16: What is this for?

db/migrations/20220723052618_dirstro_events_packing_list_entries.php
```diff
@@ -0,0 +1,85 @@
...
+use Phinx\Migration\AbstractMigration;
+
+class DirstroEventsPackingListEntries extends AbstractMigration
+{
+    public function change()
+    {
+        $distroEventsPackingListEntries = $this->table('distro_events_packing_list_entries', [
+            'id' => true,
+            'primary_key' => ['id'],
+            'engine' => 'InnoDB',
```
HaGuesto, line 12: Same comment as before

HaGuesto (changes requested):
@vahidbazzaz general comments:
- if u are setting NULL to false for a column, please define what the default value should be.
- what are all these indices for? I would only create an index if u filter or join by a column. 
- Also does it not automatically create an index if u introduce a foreign key?

db/migrations/20220723052618_dirstro_events_packing_list_entries.php
vahidbazzaz: @HaGuesto I have added this to ensure our configuration matches the peewee in the new app, but if it needs to be changed, please feel free to do so.

db/migrations/20220722085453_dirstro_events_table.php
vahidbazzaz: @HaGuesto  I have added them based on the ref docs, please refer to docs and suggest your changes as needed

db/migrations/20220722085453_dirstro_events_table.php
vahidbazzaz: @HaGuesto  In this case, the value should never be null, so why do you think there should be a default value?

db/migrations/20220722085453_dirstro_events_table.php
vahidbazzaz: @HaGuesto  the value should never be null, so why do you think there should be a default value?

db/migrations/20220723052618_dirstro_events_packing_list_entries.php
vahidbazzaz: @HaGuesto This is not so important, but make sure this matches peewee's configuration.

db/migrations/20220723052618_dirstro_events_packing_list_entries.php
vahidbazzaz: @HaGuesto also same answer for this

db/migrations/20220722085453_dirstro_events_table.php
vahidbazzaz: @HaGuesto  why do you think there should be a default value? this is a mandatory field

vahidbazzaz:
@HaGuesto Thanks for your review, please find my answers to your comments:
- The encoding, collations, and engine were removed from table definitions, but please note that this was the norm in previous migrations!
- In my opinion, mandatory fields should not always have default values. It would be helpful if you could provide your changes to the reference documents if you believe there is a case.
- I removed the extra indexes for foreign key fields per your suggestion, but kept the ones that can be filtered out in queries

db/migrations/20220722085453_dirstro_events_table.php
pylipp: FYI `utf8` is a deprecated alias in MySQL 8.0

https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-sets.html

vahidbazzaz:
@pylipp thanks for your comment regarding the `utf8` encoding, I have checked the mysql docs again and while it says that `utf8` is deprecated but it also mentioned it will become an alias for `utf8mb4` in the future:

`utf8: An alias for utf8mb3. In MySQL 8.0, this alias is deprecated; use utf8mb4 instead. utf8 is expected in a future release to become an alias for utf8mb4.`

aerinsol:
I don't agree with not putting in indices (indexes). It's not like putting
in indices is very expensive and if you already know there's going to be a
lot of filtering / queries run against a table that aren't joined or
retrieved by the PK, it makes sense.

HaGuesto:
@Vahid:
- you are correct about the defaults. I just did a quick first review from my phone and did not check the ref. Sorry for that.
- I will check what the right way is regarding regarding collation, encoding and engine.

@aerinsol:
Regarding the indices: I agree to have indices on the FOREIGN KEY columns, but  they are created automatically for a column if you add a foreign key constraint. You can define indices yourself, but this is obsolete if you add a FOREIGN KEY constraint for the same column. So, this is is not a important point. WE can explicitly mention the indices in the migration or we can leave them out. I do not have a strong opinion

[For reference](https://dev.mysql.com/doc/refman/8.0/en/constraint-foreign-key.html)
> MySQL requires that foreign key columns be indexed; if you create a table with a foreign key constraint but no index on a given column, an index is created.

HaGuesto:
@vahidbazzaz one change request I have:
can u please do two things around the `created_by` column:
- allow NULL
- make the DELETE case in the FOREIGN KEY CONSTRAINT `SET_NULL`

For GDPR reasons we need to be able to fully delete a user in our database and their references in the data. We could solve this by changing this reference to go to another user, but we do not do it with any other created_by column in all other tables and I would like to keep it that way.

db/migrations/20220723052618_dirstro_events_packing_list_entries.php
```diff
@@ -0,0 +1,83 @@
...
+                'null' => true,
+                'after' => 'created_by',
+            ])
+            ->addColumn('modified_by', 'integer', [
+                'null' => true,
+                'signed' => false,
+                'after' => 'modified_on',
+            ])
+            ->addForeignKey('distro_event_id', 'distro_events', 'id', [
+                'update' => 'CASCADE',
```
HaGuesto, line 63: @vahidbazzaz Why did you not add an delete case here?

db/migrations/20220723060012_dirstro_events_unboxed_item_collections.php
```diff
@@ -0,0 +1,99 @@
...
+                'null' => true,
+                'after' => 'created_by',
+            ])
+            ->addColumn('modified_by', 'integer', [
+                'null' => true,
+                'signed' => false,
+                'after' => 'modified_on',
+            ])
+            ->addForeignKey('distro_event_id', 'distro_events', 'id', [
+                'update' => 'CASCADE',
```
HaGuesto, line 74: same here?

db/migrations/20220723052618_dirstro_events_packing_list_entries.php
vahidbazzaz: @HaGuesto because we don't want to have soft delete for this table as we discussed with @spaudanjo

db/migrations/20220723061731_dirstro_events_outflow_logs.php
```diff
@@ -0,0 +1,94 @@
...
+                'null' => true,
+                'after' => 'created_by',
+            ])
+            ->addColumn('modified_by', 'integer', [
+                'null' => true,
+                'signed' => false,
+                'after' => 'modified_on',
+            ])
+            ->addForeignKey('distro_event_id', 'distro_events', 'id', [
+                'update' => 'CASCADE',
```
HaGuesto, line 71: and here

db/migrations/20220723060012_dirstro_events_unboxed_item_collections.php
vahidbazzaz: @HaGuesto same here, please refer to docs

db/migrations/20220723052618_dirstro_events_packing_list_entries.php
HaGuesto: Then u need to set delete to CASCADE. because if you do not define it it actually works like a RESTRICT

db/migrations/20220723060012_dirstro_events_unboxed_item_collections.php
vahidbazzaz: @HaGuesto same here

db/migrations/20220723052618_dirstro_events_packing_list_entries.php
vahidbazzaz: @HaGuesto ok sure

HaGuesto (approved):
# functional review
I went through the ref doc and checked if all fields were created correctly.

## locations
- [x]  `type` varchar(20) NOT NULL DEFAULT 'Warehouse’,^M
- [x]  `latitude` double DEFAULT NULL,^M
- [x]  `longitude` double DEFAULT NULL,^M
- [x]  `description` varchar(255) DEFAULT NULL,^M
- [x] rollback

## stock

- [x] `distro_event_id` int unsigned DEFAULT NULL,
- [x] foreign key
- [x] rollback

## distro_events

- [x]  `id` int unsigned NOT NULL AUTO_INCREMENT,^M
- [x]`name` varchar(255) DEFAULT NULL,^M
- [x]  `planned_start_date_time` datetime NOT NULL,^M
- [x] `planned_end_date_time` datetime NOT NULL,^M
- [x]  `location_id` int unsigned NOT NULL,^M
- [x]  `state` varchar(255) NOT NULL DEFAULT 'Planning',^M
- [x]  `created_on` datetime NOT NULL,^M
- [x]  `created_by` int unsigned DEFAULT NULL,^M
- [x]  `deleted_on` datetime DEFAULT NULL,^M
- [x]  `modified_on` datetime DEFAULT NULL,^M
- [x]  `modified_by` int unsigned DEFAULT NULL,^M
- [x] foreign key
- [x] rollback

## distro_event_packing_list_entries

- [x]  `id` int unsigned NOT NULL AUTO_INCREMENT,^M
- [x]  `product_id` int unsigned NOT NULL,^M
- [x]  `number_of_items` int NOT NULL,^M
- [x]  `size_id` int unsigned NOT NULL,^M
- [x]  `distro_event_id` int unsigned NOT NULL,^M
- [x]  `state` varchar(255) NOT NULL,^M
- [x]  `created_on` datetime DEFAULT NULL,^M
- [x]  `created_by` int unsigned DEFAULT NULL,^M
- [x]  `modified_on` datetime DEFAULT NULL,^M
- [x]  `modified_by` int unsigned DEFAULT NULL,^M
- [x] foreign key
- [x] rollback

## distro_events_unboxed_item_collections` (

- [x]  `id` int unsigned NOT NULL AUTO_INCREMENT,
- [x]  `product_id` int unsigned NOT NULL,
- [x]  `size_id` int unsigned NOT NULL,
- [x]  `number_of_items` int NOT NULL,
- [x]  `distro_event_id` int unsigned NOT NULL,
- [x] `location_id` int unsigned NOT NULL,
- [x]  `comments` text DEFAULT NULL,
- [x]  `created` datetime DEFAULT NULL,
- [x]  `created_by` int unsigned DEFAULT NULL,
- [x]  `modified_on` datetime DEFAULT NULL,
- [x]  `modified_by` int unsigned DEFAULT NULL,
- [x]  `box_state_id` int unsigned NOT NULL DEFAULT '1',
- [x] foreign key
- [x] rollback

## distro_events_outflow_logs

- [x]  `id` int unsigned NOT NULL AUTO_INCREMENT,
- [x]   `product_id` int unsigned NOT NULL,
- [x]  `size_id` int unsigned NOT NULL,
- [x]  `number_of_items` int NOT NULL,
- [x]  `date` datetime NOT NULL,	
- [x]  `location_id` int unsigned NOT NULL,
- [x]  `distro_event_id` int unsigned DEFAULT NULL,
- [x]  `details` text DEFAULT NULL
- [x] foreign key
- [x] rollback
- I removed the created and modified fiels since they are not needed for logs and not in the ref document

HaGuesto (approved):
@vahidbazzaz thanks for the last changes

code review is now also approved

PR #508, opened 2022-10-06
Unblock the limitation for 500 beneficiaries if the user wants

vahidbazzaz:
The PR introduced a "Page size" filter that initially shows 500 records as well as a quick filter for today's registered beneficiaries


vahidbazzaz:
[Feature Request: Unblock the limitation for 500 beneficiaries if the user wants](https://trello.com/c/zSYc9eN5)

PR #511, opened 2022-10-19
drop limit of max QR-Codes to 250

HaGuesto:


HaGuesto:
[QR generation fails in old app for bulk creation requests](https://trello.com/c/5H7ByALh/773-qr-generation-fails-in-old-app-for-bulk-creation-requests)

PR #515, opened 2022-11-30
Format state change in box history

pylipp:
Follow-up to boxwise/boxtribute#519

Currently the box history for a state change shows a message like

```
admin changed 0 from 5 to 2
```

which is confusing. This PR sets the correct change name and replaces box state IDs by labels.


HaGuesto (approved):

HaGuesto:
@pylipp first dropapp PR ever?
Congrats

PR #520, opened 2022-12-19
Fix/archive locations without instock boxes

HaGuesto:


HaGuesto:
[archive only locations which do not have instock Boxes.](https://trello.com/c/oB92x0Ln/1050-archive-only-locations-which-do-not-have-instock-boxes)

PR #527, opened 2023-02-01
Edit Warehouse Locations Explanation Text

aerinsol:
Decided that it was strange that we offer "Scrap" as a location status but don't explain, so I added text for this. Also bulleted the list of different location types. 

Signed-off-by: Roanna K <6267028+aerinsol@users.noreply.github.com>


pylipp:
@aerinsol I thought we discourage the use of SCRAP locations but encourage to update boxes' status as scrap?

aerinsol:
@pylipp updated description text. Now still retains the explanation, but is written in a passive voice so that it doesn't sound like we're encouraging them to do it.

pylipp (approved):

PR #531, opened 2023-02-22
Add cron job for daily checking expired transfer agreements

pylipp:
Not sure if DB updates work like that out of a cron job. I tested the command itself locally and it works.

Note that the `terminated_by` field will stay NULL.

The DB update might return something like the following, should/can we log that?

```
Query OK, 1 row affected (0.02 sec)
Rows matched: 1  Changed: 1  Warnings: 0
```

### DIFF
--- a/cron.yaml
+++ b/cron.yaml
@@ -28,6 +28,10 @@ cron:
   url: /cron/reseed-auth0.php
   target: staging
   schedule: every day 00:45
+- description: "check expired transfer agreements"
+  url: /cron/dailyroutine.php?action=check_expired_transfer_agreements
+  target: staging
+  schedule: every day 00:05
 
 # demo
 - description: "auth0 dailyroutine demo - check inactive users"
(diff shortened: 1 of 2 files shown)


pylipp:
[Add cron job to daily check expired transfer agreements](https://trello.com/c/liAJ9veL/1144-add-cron-job-to-daily-check-expired-transfer-agreements)

HaGuesto (approved):
@pylipp code and functional review passed :heavy_check_mark: 

Is there a reason why you implement it here and not in v2?

pylipp:
> @pylipp code and functional review passed heavy_check_mark
> 
> Is there a reason why you implement it here and not in v2?

I was not sure if it's possible to have multiple cron configs per project (similarly to dispatch config). Reading [the docs](https://cloud.google.com/appengine/docs/flexible/scheduling-jobs-with-cron-yaml) now I think it works.
I was a bit lazy to set up the cron config and CI and the endpoint for v2 :grimacing: looked like it'd work faster using dropapp but obviously at some point the cron jobs have to be moved to v2 anyways.

PR #532, opened 2023-02-24
Revert "Switch session storage to DataStore"

jamescrowley:
Reverting due to unrelated build error


HaGuesto:
[Moving Session Storage to GCloud DataStore](https://trello.com/c/xAglnJW2/1173-moving-session-storage-to-gcloud-datastore)

PR #548, opened 2023-05-11
Add Special Endpoint for V2 App Logout

vahidbazzaz:
This PR adds a special Endpoint for v2 logouts that solves the issue of v2 apps not logging out DropApp.

### DIFF
--- a/index.php
+++ b/index.php
@@ -27,6 +27,10 @@ Tracer::inSpan(
             loginWithRedirect();
         }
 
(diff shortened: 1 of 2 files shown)


vahidbazzaz:
[IMPORTANT BUG: Logout from V2 does not create a corresponding logout from Dropapp](https://trello.com/c/sbIJYHFF)

HaGuesto (approved):

PR #554, opened 2023-06-20
DB migration to add source/target quantity fields for shipment_detail table

pylipp:
- both fields are signed (like stock.items)
- target quantity is null when creating the shipment detail


pylipp:
[DB migration for source/target quantityfields in shipment detail table](https://trello.com/c/oAr1xuZx/1229-db-migration-for-source-target-quantityfields-in-shipment-detail-table)

HaGuesto (approved):
thanks @pylipp

PR #555, opened 2023-06-22
DB migration for NotDelivered box state

pylipp:


HaGuesto (approved):
Thanks @pylipp! It passed both reviews

PR #577, opened 2024-02-23
Update PHP Version From 7.4 to 8

vahidbazzaz:


library/lib/errorhandling.php
```diff
@@ -194,7 +194,7 @@ public static function add_error_handler()
      *
      * @return bool
      */
-    public static function boxwise_error_handler($errno, $errstr, $errfile, $errline, $errcontext)
+    public static function boxwise_error_handler($errno, $errstr, $errfile, $errline, $errcontext = [])
```
HaGuesto, line 200: why does this now need a default value?

library/lib/errorhandling.php
```diff
@@ -214,7 +214,6 @@ public static function boxwise_error_handler($errno, $errstr, $errfile, $errline
 }
 
 // Set error handlers
-smarty::muteExpectedErrors();
```
HaGuesto, line 217: I'm a bit afraid what will happen to our error_handling, but let's see.

db/migrations/20220205084503_add_cms_user_groups_roles.php
```diff
@@ -22,14 +22,14 @@ public function up()
...
-    public function down()
+    public function down(): void
     {
         $table = $this->table('cms_usergroups_roles');
         $table->removeIndex(['cms_usergroups_id'])
             ->removeIndex(['cms_usergroups_id', 'auth0_role_id'])
             ->dropForeignKey('cms_usergroups_id')
             ->save()
         ;
-        $table->drop('cms_usergroups_roles')->save();
```
HaGuesto, line 33: can you please exclude db migrations from the reactor refactorings? This should really not happen.

library/lib/smarty.php
```diff
@@ -30,19 +30,15 @@ public function display($template = null, $cache_id = null, $compile_id = null,
     {
         return Tracer::inSpan(
             ['name' => 'smarty:display:'.$template],
-            function () use ($template, $cache_id, $compile_id, $parent) {
```
HaGuesto, line 33: not sure if this will still work then. We should test it.

HaGuesto (commented):
I will do some small changes so it can be merged.

library/lib/errorhandling.php
vahidbazzaz: @HaGuesto I noticed this method is outdated, so I'll remove it, but I'll be sure to add the updated version in my next PR.

db/migrations/20220205084503_add_cms_user_groups_roles.php
vahidbazzaz: I agree with you, but if a feature is deprecated, we'll need to make sure the change is reflected in the migration too.

library/lib/formhandler.php
```diff
@@ -2,6 +2,13 @@
 
 class formHandler
 {
+    public $id;
```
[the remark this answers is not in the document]
vahidbazzaz: @HaGuesto At first, I set them as private, but ran into errors for some classes, so I made them all public. But, you're totally right about the formHandler, we can totally make that private.

library/lib/smarty.php
vahidbazzaz: https://wiki.php.net/rfc/arrow_functions_v2

PR #579, opened 2024-03-04
Add moved-boxes SQL query for annual report

pylipp:


aerinsol:
as discussed with @HaGuesto, this needs to be revised to cover

- Boxes created in donated locations (although tbh, this is arguable)
- Boxed created before 2023 but moved after 2023
- Boxes from archived locations.

aerinsol (changes requested):
As mentioned in the PR, needs corrections for accuracy.

pylipp:
[2.0 Improve moved-boxes statistics](https://trello.com/c/7b4VD48d/1386-20-improve-moved-boxes-statistics)

db/queries/annual_report.sql
```diff
@@ -105,9 +94,9 @@ BoxHistory AS (
         h.to_int,
         h.id AS id
     FROM history h
-    JOIN ValidBoxes s ON h.record_id = s.id AND h.to_int IS NOT null
-    AND h.id >= 1324559 AND h.tablename = 'stock'
-    AND h.changedate < "2024-01-01"
+    JOIN ValidBoxes s ON h.record_id = s.id
+    AND ((h.to_int IS NOT null AND h.id >= 1324559) OR h.changes = "Record created")
```
aerinsol, line 98: remind me what is `from_int` and `to_int` in the `history` table again?  @pylipp

pylipp: E.g. for a change box state, `changes` is `box_state_id`, and `from_int` / `to_int` the old/new value of the field.
On the opposite, for a change in a non-FK field like `comments`, `from_int` / `to_int` are null, and the changes are tracked in the `changes` field entirely (I'd prefer a `from_str`/`to_str` field at some point...).

Hence this condition selects all history entries related to FK-field (product, size, location) changes newer than 2023 and changes of number of items , and all "Record created" entries

db/queries/annual_report.sql
```diff
@@ -87,3 +87,309 @@ LEFT JOIN camps c ON p.camp_id=c.id
...
+    p.category_id,
+    TRIM(LOWER(p.name)) AS product_name,
+    p.gender_id AS gender,
+    d.source_size_id AS size_id,
+    GROUP_CONCAT(DISTINCT tr.tag_id) AS tag_ids,
+    c.name AS target_id,
+    c.id AS base_id,
+    c.name as base_name,
+    o.label as org_name,
+    "Shipment" AS target_type,
```
HaGuesto, line 343: @pylipp This should follow the target_id column. Otherwise the union does not work and the results from this query are not in the same order as the other queries.

HaGuesto (approved):

HaGuesto:
@aerinsol can you merge it when you think it is ready?

aerinsol (changes requested):
So, I have some concerns about the optimization of the CTEs as we start to process more data, especially as it appears that three of the CTEs are simply subqueries to create the BoxStateChangeVersions and CreateDonateBoxes temporary tables. I'd have to check the explain plan to be sure, but I imagine properly materilaizing the final "temporary" tables for re-use, and then dropping the previous CTEs from memory would scale much better. 

While I understand this is simply for the annual report and thus infrequent, I'm concernd that these optimization faults persist in StatViz . Some clarity in that department, and commented notes in the relevant code would be much appreciated.

aerinsol:
@pylipp if you could add the comments in the relevant areas above and confirm if it does or does not affect the StatViz queries, that would be most helpful. I imagine some of this is also @vahidbazzaz's work, so tagging him for visibility.

pylipp:
@aerinsol note added; the query here is almost identical to the one used for the moved-boxes statistic (here also base and org name info are selected), so I suppose statviz is affected to some degree

aerinsol (approved):

PR #587, opened 2024-05-02
Re-enable minimal reseeding of staging

pylipp:


pylipp:
Closed because `back/init.sql` (which was orignially created from the seed scripts here) serves as base seed in v2.

PR #588, opened 2024-05-02
Remove cron.yaml

pylipp:
Moved to system-management repo to deploy single file for dropapp and
v2 cron jobs (otherwise deploying in either app overrides settings
deployed by other app)


pylipp:
[Move cron.yaml to system-management repo](https://trello.com/c/IuMVpdSo/1403-move-cronyaml-to-system-management-repo)

PR #597, opened 2024-06-27
Comment out opencensus tracing

HaGuesto:
@vahidbazzaz I only created this PR in order to anticipate if opencensus is not working all of the sudden and dropapp is down while I am on holiday. Please only merge and deploy to production if it is needed to act quickly.

I ran the cypress tests locally and they passed. 
If we need to act since dropapp is down, please still check if the tests are running through on staging.


HaGuesto:
[dropapp: migrate/upgrade tracing away from opencensus](https://trello.com/c/d6Zab8dJ/1434-dropapp-migrate-upgrade-tracing-away-from-opencensus)

PR #603, opened 2024-07-26
Fix Incorrect Icons in iPhone & Android

vahidbazzaz:
The PR addresses the issue of the incorrect icon when creating a shortcut link from Dropapp on iPhone by updating the apple touch icon. As part of this PR, I also noticed that Android icons were not present, so I added those as well to ensure proper display on all devices.


vahidbazzaz:
[Incorrect Favicon - When creating a shortcut link on iPhone, it still uses the old Boxwise Favicon instead of the new Boxtribute icon](https://trello.com/c/glN22h37/1452-incorrect-favicon-when-creating-a-shortcut-link-on-iphone-it-still-uses-the-old-boxwise-favicon-instead-of-the-new-boxtribute-ic)

aerinsol (approved):

PR #604, opened 2024-07-29
Handle boxtribute_god role in isUserInSyncWithAuth0 check

pylipp:
This will fix errors of the `dailyroutine.php?action=auth0_active_users` cron job like 

```
PHP error: Roles of user with id 6 is out of sync between DB and Auth0.
Base ID god, is associated in Auth0 by the Roles, but not in DB
```

The issue was that the script assumed all roles to be of form `base_X_coordinator` and tried to compare the parsed base IDs and the one's that the user has assigned via their usergroup in `cms_usergroups_camps`. For the `boxtribute_god` role, the naming is different; and for the corresponding usergroup, no camps are assigned.

Until the next dropapp deploy, these changes won't have any effect.

### DIFF
--- a/library/lib/auth0.php
+++ b/library/lib/auth0.php
@@ -270,6 +270,9 @@ function isUserInSyncWithAuth0($userId)
         }
 
         foreach ($auth0UserRoles as $auth0UserRole) {
+            if ('boxtribute_god' == $auth0UserRole['name']) {
+                continue;
+            }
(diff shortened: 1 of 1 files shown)


pylipp:
[Check failing production cron jobs](https://trello.com/c/lQCpyPpR/1417-check-failing-production-cron-jobs)

vahidbazzaz (approved):
@pylipp thanks, seems fine.

PR #606, opened 2024-08-08
Use correct library-related ABPs

pylipp:
acc. to Permissions doc.

Auth0 dev-tenant is already updated.


pylipp:
[Sync between Permissions doc and dynamic permissions script](https://trello.com/c/pdATEMAL/1339-sync-between-permissions-doc-and-dynamic-permissions-script)

vahidbazzaz (approved):
@pylipp LGTM! thanks for updating the doc as well!

PR #612, opened 2024-08-27
Add DB migration to clean history transactions

pylipp:
Merge #613 first

### DIFF
--- a/db/migrations/20240827120508_clean_history_transactions.php
+++ b/db/migrations/20240827120508_clean_history_transactions.php
@@ -1,0 +1,17 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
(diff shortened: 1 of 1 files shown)


pylipp:
[dropapp Free Shop transactions are saved in the history table](https://trello.com/c/eYI74wIy/1470-dropapp-free-shop-transactions-are-saved-in-the-history-table)

vahidbazzaz (approved):
@pylipp, I checked out the PR, and everything looks good! The migration is working as expected and successfully deletes all records in the history table with the table value as `transactions`. Thanks!

pylipp:
### Functional test
I checked the staging DB after merging the PR and the query
```
select count(1) from history h 
where tablename = "transactions";
```
returns 0.

PR #615, opened 2024-09-06
Update handling of special users

pylipp:
- god users don't have any active bases but name, email, usergroup
  should still be in sync
- some users don't have any active bases but are not marked as deleted (e.g. ID 14)
- move definition of validationResult
- move computation of false_key

This was found because the cron-job for active users still complains
about users being out of sync with the message

    PHP error: User with id 13 is out of sync between DB and Auth0  in
    /workspace/library/lib/auth0.php on line 280

but doesn't indicate any invalid key.

### DIFF
--- a/library/lib/auth0.php
+++ b/library/lib/auth0.php
@@ -222,12 +222,11 @@ function isUserInSyncWithAuth0($userId)
 
     $auth0User = getAuth0User($userId);
     $hasActiveBase = (sizeof($dbUser['active_base_ids']) > 0);
+    $validationResult = [];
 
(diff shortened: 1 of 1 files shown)


vahidbazzaz (approved):
@pylipp Thanks a bunch for fixing this! I tested it locally and your changes works fine.

pylipp:
[dropapp: Update handling of special users in isUserInSyncWithAuth0](https://trello.com/c/cggQ9HRf/1515-dropapp-update-handling-of-special-users-in-isuserinsyncwithauth0)

PR #618, opened 2024-09-09
Use Auth0 role for boxtribute_god user, too

pylipp:
Follow-up to #604


pylipp:
[dropapp: Update handling of special users in isUserInSyncWithAuth0](https://trello.com/c/cggQ9HRf/1515-dropapp-update-handling-of-special-users-in-isuserinsyncwithauth0)

PR #622, opened 2024-09-30
Null-size migration for measure products

pylipp:
- Add migrations for new sizegroups, units, and stock unit/value
- Add DB migrations to make size fields nullable

### DIFF
--- a/db/migrations/20240930122935_make_stock_size_nullable.php
+++ b/db/migrations/20240930122935_make_stock_size_nullable.php
@@ -7,7 +7,7 @@ class MakeStockSizeNullable extends AbstractMigration
     public function up(): void
     {
         $stock = $this->table('stock');
-        $stock->changeColumn('size_id', 'integer', ['null' => true])
+        $stock->changeColumn('size_id', 'integer', ['null' => true, 'signed' => false])
             ->update()
         ;
     }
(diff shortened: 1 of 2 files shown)


db/migrations/20240930122935_make_stock_size_nullable.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+use Phinx\Migration\AbstractMigration;
+
+class MakeStockSizeNullable extends AbstractMigration
+{
+    public function up(): void
+    {
+        $stock = $this->table('stock');
+        $stock->changeColumn('size_id', 'integer', ['null' => true])
```
HaGuesto, line 10: note to myself: check if FK is complaining

HaGuesto (commented):
# code review passed

HaGuesto:
# functional review
@pylipp in the changecolumn command if you do not keep the signed option in there it will just drop it and default to 'signed' -> true.

On the down migration the FK complained that there is no translation of NULL to any other value. If we really have to down migrate and boxes or shipment details exist with NULL size_id we will have to put a value there manually.

Otherwise the migration passes no and I will merge it.

HaGuesto:
@pylipp can you assign a trello ticket for this PR?

PR #624, opened 2024-10-10
null size migrations were wrongly merged in

HaGuesto:


pylipp:
thx for fixing @HaGuesto !

PR #632, opened 2024-12-08
Clarifying that "In Shop?" is actually to pin a product in the stockroom table

aerinsol:
See discussion here: https://boxwise.slack.com/archives/CT4SNP5B2/p1732498410328479

PR #633, opened 2024-12-11
Update products_edit.php

aerinsol:
clarifies the "in shop?" functionality and what "always in stockroom"means


aerinsol:
Yep, it's here https://github.com/boxwise/dropapp/pull/632

vahidbazzaz (approved):

PR #652, opened 2025-01-27
DB migration for shareable-link table

pylipp:
open questions:
1. `code` column as fixed-length char type?
2. `url_parameters` column max-length (65535) sufficient?
3. extra datetime column `last_accessed_on` and integer column `number_of_accesses` to keep track of access, and have some means for basic rate-/access-limiting (then again, these would require DB-write access for the public endpoint, not sure if we want that)?

Answers:
1. not important for now
2. use max-length 2000
3. for rate-limiting we'd use Google Armor instead of doing a software/DB solution. For tracking number of accesses internally we can use heap, for externally tracking it we might want to have an extra table at some point.

### DIFF
--- a/db/migrations/20250127114447_add_shareable_link_table.php
+++ b/db/migrations/20250127114447_add_shareable_link_table.php
@@ -33,24 +33,29 @@ class AddShareableLinkTable extends AbstractMigration
             ])
             ->addColumn('base_id', 'integer', [
                 'null' => true,
+                'signed' => false,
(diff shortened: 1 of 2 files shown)


pylipp:
[dropapp: DB migration for new shareable-link table](https://trello.com/c/0jXEacsc/1664-dropapp-db-migration-for-new-shareable-link-table)

HaGuesto (approved):

PR #674, opened 2025-03-05
Hide None-gender label in most-popular-item name on start page

pylipp:
Maybe more efficient to do

```
select
	SUM(count) as count,
	g.id as gender_id,
	p.name as product_name
FROM transactions t
        ...
```
instead, and then the name combination logic, deepening on `gender_id`, when assigning `$data['popularname']`.


pylipp:
https://trello.com/c/wyEV69DA

vahidbazzaz:
@pylipp @HaGuesto I just fixed two more cases where the text was showing as -1 or none for popular items.

Screenshot 2025-03-06 at 8 22 13 PM

Screenshot 2025-03-06 at 8 31 00 PM

PR #676, opened 2025-03-07
show both product tabs only if beta level agrees

HaGuesto:


HaGuesto:
[dropapp: Hide StandardProducts tab for users with beta-level < 4](https://trello.com/c/sTm0aEG1/1694-dropapp-hide-standardproducts-tab-for-users-with-beta-level-4)

PR #695, opened 2025-04-24
Update rolesToActions

pylipp:
@haguesto should the "Services" menu be added to the `menusToActions` list?


library/constants.php
```diff
@@ -373,6 +373,45 @@
             'view_transfer_agreements',
         ],
     ],
+    [
+        'id' => 168,
+        'module' => 'Sevices',
```
pylipp, line 378: small typo @HaGuesto

PR #739, opened 2025-08-21
Upgrade composer packages

HaGuesto:
FYI @vahidbazzaz @pylipp 
- removed unused package
- tried to upgrade all packages except smarty and the qr generator
- I did not do any major version upgrade besides sendgrid
- unfortunately the deprecated opencensus package causes conflicts regarding the sub dependency `psr/log`. Therefore, I downgraded some other packages. So, the current version numbers are most likely the highest for some packages that are possible with opencesus still in the composer.json
- I ran all cypress tests locally and they run through.


HaGuesto:
my local php and the build php version in circleci are not in sync. I'm fixing it atm.

PR #766, opened 2025-10-06
remove logging qr code generation to history

pylipp:
https://trello.com/c/70vc78uU

- Don't log QR code generation to history table
- Add comments
- Remove handling of legacy QR code
- Correctly use simpleBulkSaveChangeHistory


pdf/qr.php
```diff
@@ -75,23 +77,17 @@ function () {
             }
 
             // qr code generation
```
vegamark061-code: Don't understand this

PR #834, opened 2026-07-25
Validate warehouse location name length instead of silently truncating it

vahidbazzaz:
## Summary
- "Edit Warehouses" let you save a location name longer than the DB column (`locations.label varchar(20)`), which MySQL silently truncated with no feedback.
- Widens `locations.label` to `varchar(50)` via phinx migration, and adds matching client-side (jQuery Validate, blocks submit) and server-side (`mb_strlen` check, throws a clear exception) validation, plus a tooltip explaining why the limit exists (mobile navigation speed for warehouse/distribution teams).
- Along the way, found and fixed two pre-existing latent bugs the change surfaced:
  - The `data-max-count` character counter's jQuery plugin (`jquery.simplyCountable.js`) existed in the repo but was never loaded on any page — no field had ever used the `maxlength` option before, so this was silently dead code.
  - The counter's label text was hardcoded in Dutch ("tekens over van") instead of using the already-seeded-but-unused English translate key `cms_form_charactersleft`.

## Test plan
- [x] Applied the migration locally and confirmed `locations.label` is `varchar(50)` via `SHOW COLUMNS`
- [x] Scripted the real login + form POST flow against the running app: a 51-char name is rejected with the new error and never reaches the DB; a 50-char name saves in full, untruncated
- [x] Added `cypress/e2e/1_feature_tests/6_1_Warehouse_Locations.js` covering both cases; ran it against the live app in a real (headless) browser — both pass
- [x] Manually confirmed in-browser: inline error + tooltip render correctly, and the counter now shows English text ("50 characters left from 50") instead of Dutch


vahidbazzaz:
[Dropapp Bug: Edit Warehouse Locations Names Are Too Short in Limitations #small](https://trello.com/c/WxJAYpQe/2073-dropapp-bug-edit-warehouse-locations-names-are-too-short-in-limitations-small)

